PR #57, opened 2020-05-13
Added support for nullable nested dicts and lists

sondrelg:


sondrelg:
Changes error messages from 

```
Summary
-------------------------------------------------------------------------------------------

Error:      Mismatched types. Expected response to be <class 'dict'> but found <class 'NoneType'>.

Expected:   {...}
Received:   None

Hint:      
Sequence:   init.dict:key:vendorOrders.list.dict:key:trackingNumber

-------------------------------------------------------------------------------------------

* If you need more details: set `verbose=True`
```

to

```
Summary
-------------------------------------------------------------------------------------------

Error:      Mismatched types. Expected response to be <class 'dict'> but found <class 'NoneType'>.

Expected:   {...}
Received:   None

Hint:       Your schema needs to set `x-nullable: True` if there's a change it will return None.
            For drf-yasg implementations, set `x_nullable=True` in your Schema definition.
Sequence:   init.dict:key:vendorOrders.list.dict:key:trackingNumber

-------------------------------------------------------------------------------------------

* If you need more details: set `verbose=True`
```

and allows correctly documented items to pass silently. The current implementation raises the first error, even when `x-nullable` is correctly documented as  `True`

PR #67, opened 2020-09-11
Version 2.0

sondrelg:
- Reworks settings so that different implementations use different `SCHEMA_LOADER`'s, rather than different import locations
- Reworks settings so that case-testing is done using a callable passed to `CASE_TESTER`
- Cuts repeat iterations by merging case checking with schema checking
- Reimplements demo project using Django 3.1
- Adds an `AppConfig`


sondrelg:
to do:
- [ ] Add changelog to CHANGELOG.rst
- [ ] Update README and PYPI_README
- [ ] Implement some backward compatibility for the `response_validation` import

sondrelg (commented):
Looks OK to me. Will have one more read through the docs to make sure everything is up to date.

PR #75, opened 2020-09-19
Updated pre-commit config

Goldziher:
This PR updates the pre-commit configs. 

I recommend running "pre-commit run --all-files" after merging it, and then applying all fixes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,14 +15,14 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies: [
-            'flake8-bugbear==19.8.0',  # Looks for likely bugs and design problems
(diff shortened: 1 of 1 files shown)


sondrelg:
Nice! Happy to merge this, and I run it on the project after.

One thing: I'm not familiar with `pyupgrade`; will the `--py37-plus` flag implement 3.7 language features? Just asking because I thinking Django 2.2 supports pyton 3.6, so I think it would be good to make sure this package can run on 3.6+ as well.

Will it break backwards compatibility for python 3.6?

Goldziher:
@sondrelg it upgrades to using language features from the given version - perhaps thatvflag should be removed then 😃

Goldziher:
done

PR #78, opened 2020-09-23
Caching

sondrelg:
Adds a cache layer to prevent duplicate validation of the same response objects.


sondrelg:
Ran a few tests on a big work project endpoint, and without caching, the whole validation process added about 65ms to *each request*. After adding caching, the schema loading adds just under 10ms to a request, and only the first response out of *n* responses is validated.

sondrelg (commented):
Looks OK, but will make sure to test the caching dynamic manually just in case.

PR #81, opened 2020-09-24
Fixes caching, adds validation-exempt-status-codes setting, adds support for viewsets, and updates models

sondrelg:


sondrelg (commented):
Looks OK

PR #89, opened 2020-10-21
Update test workflow

sondrelg:
### Changes

- CI test stage adds extra django and python version to test matrix, adds a separate linting job, and implements caching.
- Ran black and isort on project with slightly tweaked settings


sondrelg (commented):
All just linting + ci-template change.

PR #117, opened 2020-12-13
feat(added django test case)

Goldziher:
This PR adds a DRF TestCase that can be extended when using Django based testing.


Goldziher:
This is in a draft state because it still needs tests. I am also not sure about adding an `assertRequest` method as well.

sondrelg:
Cool! This is what I wanted to do when starting the project, but moved away from it because I didn't see an obvious way of getting the method and route in a reliable way. This looks promising however 👍 

If this works, do you think it would be possible to create a wrapper function to automatically apply testing to individual tests as well? This way you might be able to test a 200-response test and a 400-response test, but leave the rest alone?

Goldziher:
ok @sondrelg PR is finished.

django_swagger_tester/testing.py
```diff
@@ -74,3 +76,16 @@ def validate_input_serializer(
...
+    """ Extended Test Case with OpenAPI assertions"""
+
+    def assertResponse(self, response: Response, **kwargs: Any) -> None:
+        """ Assert response match OpenAPI spec """
+        route = kwargs.pop('route', response.request['PATH_INFO'])
+        method = kwargs.pop('method', response.request['REQUEST_METHOD'])
+        try:
+            validate_response(response=response, method=method, route=route, **kwargs)
+        except (SwaggerDocumentationError, CaseError) as e:
+            raise self.failureException from e
```
sondrelg, line 91:
```suggestion
class OpenAPITestCase(APITestCase):
    """
    Extends APITestCase with OpenAPI assertions.
    """

    def assertResponse(self, response: Response, **kwargs: Any) -> None:
        """
        Assert response matches the OpenAPI spec.
        """
        route = kwargs.pop('route', response.request['PATH_INFO'])
        method = kwargs.pop('method', response.request['REQUEST_METHOD'])
        try:
            validate_response(response=response, method=method, route=route, **kwargs)
        except (SwaggerDocumentationError, CaseError) as e:
            raise self.failureException from e
```

sondrelg: ^ Just to keep styling consistent.

sondrelg (changes requested):
Please delete the `.DS_Store` file; otherwise I have no comments, just a suggested docstring change for consistency.

Also, if you didn't see it, see my question about wrapper functions!

Good work 🚀

Goldziher:
Removed DS_Store, 

I didnt see your question. What is it?

> If this works, do you think it would be possible to create a wrapper function to automatically apply testing to individual tests as well? This way you might be able to test a 200-response test and a 400-response test, but leave the rest alone?

Well for django tests it shouldnt be a problem - its basically a decorator. You can also do this with the client. 

On the other hand - my opinion on testing best practices is that tests should not obfuscate implementation: In production code its good to keep stuff dry. In tests though, repetition is fine because a good test is readable and descriptive, and above all - easy to maintain by people other than the person(s) who wrote it. 

So adding this kind of wrapper (not very pythonic - more JS tbh) or a decorator is possible, but I would avoid doing this.

sondrelg (approved):
🎉

Issue #115, opened 2020-12-10
Path parameters and i18n


saeub:
We have an internationalized API where the language is provided in the path like this:

- `/de/api/objects/{id}`
- `/en/api/objects/{id}`

It would be nice to document this in the OpenAPI like this:

```yaml
paths:
  /{lang}/api/objects/{id}:
    get:
      parameters:
        - name: lang
          in: path
          ...
```

However, since we're using `django.conf.urls.i18n.i18n_patterns` to generate the URL patterns for all available languages, `lang` isn't an actual parameter in the Django URL pattern. `django_swagger_tester` seems to assume that all parameters in the YAML schema are also parameters in the Django URL pattern and raises an exception:

```
Error: Unsuccessfully tried to index the OpenAPI schema by `/de/api/objects/123`.

For debugging purposes, other valid routes include: /{lang}/api/objects, /{lang}/api/objects/{id}.
```

The only solution I found to this is to hard-code all available languages in the schema:

```yaml
paths:
  /de/api/objects/{id}:
    ...
  /en/api/objects/{id}:
    ...
  ...
```

But this is obviously not very nice, since the response schemas always look the same. Is there another way to do this? Did I miss something?

For example, would it possible for us to tell `validate_response()` which value to expect for the additional `lang` parameter in the schema and add it to the [`kwargs`](https://github.com/snok/django-swagger-tester/blob/6b70869216adf790025c793570d91e88152564db/django_swagger_tester/utils.py#L188) that were returned by Django's `resolve()` function, so it can resolve `/de/api/objects/{id}` into `/{lang}/api/objects/{id}` to get the correct route for indexing the schema?

Thanks!

sondrelg:
This is interesting. I agree you should be able to document the endpoint like your first example.

Do you think internationalised routes is something that could be auto-detected? **If** we had reason to believe that everyone would document their endpoints like `/{lang}/api/objects/{id}` (I wonder how swagger-generators behave), I would think that this is something we could default to. If not, maybe we could create a setting to treat the locale-part of the route as a parameter.

Would you be willing to submit a PR with the minimal amount of code required to set up similar paths in the demo project?

saeub:
Thanks for the quick reply @sondrelg!

I'm not entirely familiar with how Django generates those routes, so I'm not sure how deep we would have to dig to detect the internationalized ones. Also, I haven't seen this pattern of using language codes in URLs to control response language used a lot in API specifications generally, and I don't know if there is a convention.

[...]

PR #123, opened 2020-12-17
Document and test i18n endpoint in static schema

saeub:
As discussed in #115.

I added the internationalized route in `demo/static_schemas/openapi-schema.yml` and added some test items in `tests/test_testers/test_validate_response/test_static.py`. This now reproduces the error I got when documenting the endpoint as `/{lang}/api/...`.

I also noticed that some tests relating to the static schemas didn't seem to work properly, so I tried to fix them (let me know if I just misunderstood them). Specifically, the `yml_path` in `test_static.py` seemed to point to the wrong YAML file (?), and the list comparison test in `test_get_endpoints.py` didn't actually check the content of the lists.

### DIFF
--- a/demo/api/swagger/auto_schemas.py
+++ b/demo/api/swagger/auto_schemas.py
@@ -167,7 +167,7 @@ def languages_auto_schema():
                 title='Success',
                 type=TYPE_OBJECT,
                 properties={
-                    'Languages': Schema(
+                    'languages': Schema(
                         title='Success',
                         type=TYPE_ARRAY,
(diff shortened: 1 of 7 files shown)


sondrelg:
Nice 👏  I'm a bit busy the next few days, but should be able to look at this over the weekend most likely! 

Thanks for the effort so far, and just in case it's not obvious, if you want to try and implement a fix in the meantime, you are of course very welcome to do so 😄

sondrelg (approved):
👍

PR #125, opened 2020-12-20
Add support for parameterizing the i18n component of a schema path

sondrelg:
## Changes

- Moved package settings validation into the app-config to avoid circular imports
- Added a setting which allows you to specify a parameter name that's inserted in place of the i18n component of a path
- Tweaked the sphinx setup to make it work on my mac, and deleted no longer used images


sondrelg:
New docs

image

sondrelg (commented):
Ready to review

sondrelg:
@saeub, I believe the feature (#125) is now working. Could you please test this PR on your project?

To test you will need to install the branch directly from github, like this: 

```
pip install git+https://github.com/snok/django-swagger-tester/tree/i18n
```

You should also make sure to add `SWAGGER_SETTINGS` to your project, where you specifically add `PARAMETERIZED_I18N_NAME` to match the name of the parameter in your schema. See the demo project settings.py if you need help with the settings structure.

---------------

After you've tested that it works, also feel free to review the code. Things like

- the setting name
- the docs
- how we've implemented the setting

are all just written in a hurry, so I would love your input on these.

If you feel like anything could be improved, don't hold back 🏄‍♂️

django_swagger_tester/configuration.py
```diff
@@ -139,47 +142,58 @@ def validate(self):
...
         self.validate_case_passlist()
         self.set_and_validate_schema_loader()
+        self.validate_parameterized_i18n_name()
         self.middleware_settings
         self.view_settings
 
+    def validate_parameterized_i18n_name(self):
+        if not isinstance(self.parameterized_i18n_name, str):
+            raise ImproperlyConfigured('PARAMETERIZED_I18N_NAME must be a string')
+
```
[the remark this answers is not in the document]
JonasKs: Nice, easy to read and understand too :+1:

JonasKs (approved):
Not much to say, it's clean and easy to read. :santa:

django_swagger_tester/utils.py
```diff
@@ -262,7 +261,8 @@ def get_path(self) -> str:
         path = f'{path[:starting_index]}{self.resolved_path.kwargs[parameter_name]}{path[starting_index + len(parameter):]}'
         self.parameterized_path = path
         self.counter += 1
-        return path
+
+        return self.replace_i18n_parameter(path)
```
saeub, line 265: `self.replace_i18n_parameter(self.parameterized_path)` also needs to be returned when `self.counter == 0` (line 253). Currently, the i18n parameter only gets replaced when there is also another parameter in the path (you can test this by removing the `version` parameter of the `/api/v1/i18n` endpoint).

saeub (changes requested):
Thanks a lot, this looks great! I found a bug in `Route.get_path()` and some mistakes in the static schema from my PR (oops!). Otherwise, it seems to be working like a charm in our project.

django_swagger_tester/utils.py
sondrelg: Good catch 👏

sondrelg:
@saeub Thanks for testing 👏  All changes are now applied. Feel free to test again if you want. Unless something else comes up I think I'll merge and release a new version tomorrow 🏄‍♂️

saeub:
@sondrelg Awesome, thank you so much! Looks like it's working :+1:

PR #132, opened 2020-12-31
Deprecations and package renaming

sondrelg:
## Deletions

  - Middleware
  - APIViews
  - All caching log, hashing logic, and migrations
  - All tests for the above logic
  - Related docs

## Renaming
- Renames the package from `django-swagger-tester` to `django-openapi-tester`
- Renames the source directory from `django_swagger_tester` to `openapi_tester`
- Renames the package settings from `SWAGGER_TESTER` to `OPENAPI_TESTER`


Goldziher:
i suggest calling this simply drf-openapi-tester or django-openapi-tester.

Either way, I am waiting for you to finish this PR so I can branch from it.

sondrelg:
@Goldziher , it's a good thing you said - I don't think you need to wait for me if you want to start, I'll resolve any conflicts on my side if they arise.

If you prefer to wait, I'll try to merge deprecations/removals in ~2 days fyi 👍

sondrelg:
Re the name, I think django-openapi-tester sounds good. Think it's possible DRF can be moved to an optional dependency, so having that in the name might be misleading.

Goldziher:
@sondrelg  - there is really no point of having openapi specs without drf :), i mean - if you are going to document a SOAP api you are probably not using django and defintely not using openapi. Otherwise drf is IMO a must.

Goldziher:
im rebasing my PR to branch from this PR - this means my PR will not be green anymore, but it will make it easier to continue to work. There is simply too much noise in the package as is and quite a bit of pasta :)

saeub:
I agree with @sondrelg that having DRF in the name is misleading, since you can implement a REST API without using DRF.

I currently have a project where I essentially have an existing OpenAPI specification and need to implement a REST API that fulfills it. Since the API is rather simple and I do not use most of the features DRF provides, I decided not to use DRF at all and just use plain Django views. I would still love to use this package for validating the responses.

Goldziher:
well its not so much about the name as about DRF being an optional dependency - I dont think this is possible, or desirable.

Goldziher:
@sondrelg my PR now points at this PR. I fixed the tests on my end, so perhaps its best if your merge mine into this one. But i leave that to you. I already fixed locally the allOf issue, i will add a followup PR with this stuff ASAP.

sondrelg (commented):
Very messy at the moment, but I think it's best to merge this now so I can rename the repository, readthedocs, and the pypi-package name. 

Looked throught everything pretty thoroughly, and anything deleted by accident remains in the git log anyways.

Issue #131, opened 2020-12-30
Rework and trim down functionality for v3 (discussion)


sondrelg:
I want to open this ticket to flag some ideas and suggestions for deprecations going forward. 

### Renaming the package & deprecating logic

Once #116 is resolved, I think it might be a good time to clearly define the purpose of this package, and (perhaps to) get rid of some code.

Unless anyone chimes in with good reasons not to, I would like to:

1. Deprecate the middleware and custom views that were introduced this summer but that I don't even use myself (so I cannot be sure it really works all that well)
2. Deprecate the input validation code
3. Rename the package from `django-swagger-tester` to `django-openapi-response-tester` or something similar

I'm assuming almost no one has used the code I'm proposing to deprecate, and if someone wanted an `APIView` with builtin response validation, this is something anyone could easily do, without it having to live in this repo.

The re-defined purpose of the package could then be to *only* provide a mechanism for validating API responses against OpenAPI schema response sections. That's basically what we do now, but I think re-branding the package would make the documentation a lot clearer.

### Leveraging third party packages

I think @Goldziher's suggestion to add third-party library for schema parsing is a good one, and I don't necessarily think we should stop at one.

Another area of the package that I think could be improved by external dependencies is the whole validation logic. Right now the validation logic is a thick mix of validation logic, parsing logic, and custom exception messages. If we instead were able to reduce the OpenAPI response schema to a plain type-definition, we should be able to completely separate the validation logic from any type of parsing logic, making it everything more concise and predictable. We might even be able to outsource the validation completely, and instead focus on outputting legible error messages.

------------------

The goal of this would be to remove complexity and to create a clearer purpose for the package. The ideal outcome (in my opinion) is to nail (just) response validation and then stop adding features 🚀

Goldziher:
Yes, I agree with all of your points. I would suggest making the response validation logic as straightforward as possible and output an informative diff in case of a failure (expected vis-a-vis the received). Staying within the paradigm of django testing / pytest.

JonasKs:
Finally got some time to have a look. Yes, I agree!😊 Happy New Years!🍾

PR #133, opened 2021-01-02
Added Prance

### LINKED ISSUE #116: Add support for loading schema with oneOf, anyOf, allOf, not $refs
Hi fellas,

I get this:

```
 File "/usr/local/lib/python3.8/site-packages/django_swagger_tester/testing.py", line 37, in validate_response
    verbose_error_message = format_response_tester_error(e, hint=e.response_hint)
  File "/usr/local/lib/python3.8/site-packages/django_swagger_tester/utils.py", line 47, in format_response_tester_error
    example_item = settings.loader_class.create_dict_from_schema(exception.schema)
  File "/usr/local/lib/python3.8/site-packages/django_swagger_tester/loaders.py", line 335, in create_dict_from_schema
    return self._iterate_schema_dict(schema)
  File "/usr/local/lib/python3.8/site-packages/django_swagger_tester/loaders.py", line 294, in _iterate_schema_dict
    elif read_type(value) == 'object':
  File "/usr/local/lib/python3.8/site-packages/django_swagger_tester/openapi.py", line 53, in read_type
    raise OpenAPISchemaError(
django_swagger_tester.exceptions.OpenAPISchemaError: Schema item has an invalid `type` attribute. The type should be a single string.
```

I then see this printed: 
`Schema item: {'allOf': [{'type': 'object', 'description': 'Read-only Product Serializer', 'properties': {'slug': {'type': 'string', 'readOnly': True, 'pattern': '^[-a-zA-Z0-9_]+$'}, 'loop': {'type': 'string', 'readOnly': True}, 'sites': {'type': 'array', 'items': {'type': 'object', 'description': 'read-only Site serializer', 'properties': {'id': {'type': 'integer', 'readOnly': True}, 'slug': {'type': 'string', 'readOnly': True, 'pattern': '^[-a-zA-Z0-9_]+$'}, 'domain': {'type': 'string', 'readOnly': True}}, 'required': ['domain', 'id', 'slug']}, 'readOnly': True}, 'acr': {'allOf': [{'enum': ['3', '4'], 'type': 'string'}], 'readOnly': True}}, 'required': ['acr', 'loop', 'sites', 'slug']}], 'readOnly': True}`

which I prettified:
```
{
  allOf: [
    {
      type: "object",
      description: "Read-only Product Serializer",
      properties: {
        slug: { type: "string", readOnly: True, pattern: "^[-a-zA-Z0-9_]+$" },
        loop: { type: "string", readOnly: True },
        sites: {
          type: "array",
          items: {
            type: "object",
            description: "read-only Site serializer",
            properties: {
              id: { type: "integer", readOnly: True },
              slug: {
                type: "string",
                readOnly: True,
                pattern: "^[-a-zA-Z0-9_]+$",
              },
              domain: { type: "string", readOnly: True },
            },
            required: ["domain", "id", "slug"],
          },
          readOnly: True,
        },
        acr: { allOf: [{ enum: ["3", "4"], type: "string" }], readOnly: True },
      },
      required: ["acr", "loop", "sites", "slug"],
    },
  ],
  readOnly: True,
};
```

I dont see any issue with the type key here.

Goldziher:
This PR adds prance to handle dereferencing.

It doesn't yet resolve #116, which will happen in a followup PR

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -11,7 +11,6 @@ ignore=
     D210
     # Imperative docstring declarations
     D401
-
     # Type annotation for `self`
     TYP101
     TYP102  # for cls
@@ -45,16 +44,5 @@ ignore=
 
     # Missing type annotation for **kwargs
     ANN003
-
-exclude =
-    .git,
-    .idea,
-    __pycache__,
-    demo/*,
-    tests/*,
-    venv,
-    manage.py,
-    .venv
-
 
 max-complexity = 16
--- a/.flake8
+++ b/.flake8
@@ -44,5 +44,11 @@ ignore=
 
     # Missing type annotation for **kwargs
     ANN003
-
+exclude =
+    .git,
+    .idea,
+    __pycache__,
+    venv,
+    manage.py,
+    .venv
 max-complexity = 16
--- a/.flake8
+++ b/.flake8
@@ -8,6 +8,7 @@ ignore=
     D101
     # Make docstrings one line if it can fit.
     D200
+    D210
     # Imperative docstring declarations
     D401
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,9 +3,9 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
-        args: ['--quiet']
+        args: ["--quiet"]
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v3.4.0
     hooks:
       - id: check-case-conflict
       - id: end-of-file-fixer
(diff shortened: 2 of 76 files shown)


.flake8
```diff
@@ -45,15 +45,4 @@ ignore=
...
 
-exclude =
-    .git,
-    .idea,
-    __pycache__,
-    demo/*,
-    tests/*,
-    venv,
-    manage.py,
-    .venv
```
sondrelg, line 56: Why did you remove this?

.pre-commit-config.yaml
```diff
@@ -3,52 +3,42 @@ repos:
...
       - id: flake8
-        additional_dependencies: [
-            'flake8-bugbear==20.1.4',  # Looks for likely bugs and design problems
-            'flake8-comprehensions==3.2.3',  # Looks for unnecessary generator functions that can be converted to list comprehensions
-            'flake8-deprecated==1.3',  # Looks for method deprecations
-            'flake8-use-fstring==1.1',  # Enforces use of f-strings over .format and %s
-            'flake8-print==3.1.4',  # Checks for print statements
-            'flake8-docstrings==1.5.0',  # Verifies that all functions/methods have docstrings
-            'flake8-type-annotations==0.1.0',  # Looks for misconfigured type annotations
-        ]
```
sondrelg, line 30: Why did you remove these?

.pre-commit-config.yaml
```diff
@@ -3,52 +3,42 @@ repos:
...
       - id: pyupgrade
         args: ["--py36-plus"]
-  - repo: https://github.com/pycqa/isort
-    rev: 5.5.2
+  - repo: https://github.com/asottile/seed-isort-config
+    rev: v2.2.0
+    hooks:
+      - id: seed-isort-config
+  - repo: https://github.com/pre-commit/mirrors-isort
+    rev: "v5.7.0"
```
sondrelg, line 30: Both of these repos say they're deprecated and that one should use `isort` directly. Unless you have a good reason to change this, I think it should keep the current config.

sondrelg (changes requested):
Looks like there's almost 4000 changes here, where the strong majority is just single-to-double quotes. Please revert the quote-changes.

.pre-commit-config.yaml
Goldziher: for the typing stuff - you already had mypy in place, no need to have flake8 do an inferior job. For the rest of it - these did not actually contribute to find any errors in the code base. If we want these type of checks we better use pylint (although its somewhat dubious as well):

.pre-commit-config.yaml
Goldziher: done

.flake8
Goldziher: because its not necessary. flake8 runs only against python files and i needed to run it against the stuff in tests.

.pre-commit-config.yaml
Goldziher: also the main reason - to enable running flake8 against the test files as well. This allowed me to find a lot of issues and fix them (especially missing and imports and unused variables etc). 

> Looks like there's almost 4000 changes here, where the strong majority is just single-to-double quotes. Please revert the quote-changes.

not possible, this is black.

sondrelg: > not possible, this is black.

No, this happened because you removed the `double-quote-string-fixer` from the pre-commit config and ran it on the project.

> for the typing stuff - you already had mypy in place, no need to have flake8 do an inferior job

What does mypy have to do with:

- Fixing quotes
- Detecting private keys (I don't really care about this hook)
- Detecting bad list comprehensions
- Bugbear
- Method deprecations 
- Checking for print statements
- Checking for docstrings
- Checking that f-strings are used over format

The only flake8-hook related to mypy is the type annotations one, and mypy does not enforce type annotations. For example, this code would pass mypy's tests without errors

```python
def test(x, y):
    return x + y
    
test(1, 3)
```

With the flake8 type-annotations hook this would not pass the pre-commit check.

-------------------------

I think the best thing would be to undo these removals and run `pre-commit run --all-files`. That should eliminate the excess changes.

Let me know if you want me to do it :)

.flake8
sondrelg: I have my virtual environments in my project directory, and venv's are propped full of python files. Think it would be better to only remove the `tests/*` line and keep everything else as it was.

Goldziher: pre-commit doesnt run against git ignored files...

as you wish though.

.pre-commit-config.yaml
Goldziher: followed your suggestion. it worked!

response_tester/type_declarations.py
```diff
@@ -0,0 +1,43 @@
...
+    Exporter.export_anything(__name__)
+
+else:
+    """
+    if we are type checking, import/define the actual classes/types
+    We don't have to worry about circular imports here, since
+    the type checker does not actually run the code.
+    """
+    # noinspection PyUnresolvedReferences
+    from response_tester.loaders import BaseSchemaLoader, StaticSchemaLoader
```
sondrelg, line 43: cool 😊

sondrelg (approved):
Nice! Pretty happy to merge this, but if this sounds OK, I think I will finalize #131, merge that, then merge this into master right after. 

Thinking I will sit down to do it tomorrow 🎉

Goldziher:
> Nice! Pretty happy to merge this, but if this sounds OK, I think I will finalize #131, merge that, then merge this into master right after.
> 
> Thinking I will sit down to do it tomorrow 🎉

of course. I will continue working on my other PR when time allows - probably not before the weekend.

PR #134, opened 2021-01-03
added schema validation

Goldziher:
This PR updates the validation done by the package by using the (only) python based openapi schema validator. 

Aside from being agreed upon in the discussion - this is necessary for handling of oneOf, allOf etc., because the preexisting validation needs to be thoroughly reworked - so its simpler to simply strip it away first. 

current state - 
ready for being merged - pointing at my previous PR


demo/api/views/cars.py
```diff
@@ -20,15 +20,15 @@ def get(request: Request, version: int, **kwargs) -> Response:
         return Response(cars, 200)
 
     @staticmethod
-    def put(request: Request, version: int, vehicle_type: str) -> Response:
```
Goldziher, line 23: this apiview is badly designed (i know the demo project comes from the drf-yasg repo) and it failed schema validation because, well, this is a list endpoint that never uses the path parameter.

sondrelg:
I've now updated the `v3` branch where I rename the source directory and the package name to `django-openapi-tester`. Unfortunately this creates an enormous amount of conflicts, so I think the best thing to do might be to re-apply the changes here (unless you're a git wizard and know of a better solution). Let me know if you need help.

I think I'll merge v3 tonight, and make any further changes in a new separate branch so we're not depending on each other.

Goldziher:
> I've now updated the `v3` branch where I rename the source directory and the package name to `django-openapi-tester`. Unfortunately this creates an enormous amount of conflicts, so I think the best thing to do might be to re-apply the changes here (unless you're a git wizard and know of a better solution). Let me know if you need help.
> 
> I think I'll merge v3 tonight, and make any further changes in a new separate branch so we're not depending on each other.

@sondrelg  yes i can resolve these, but you need to revert your merge to master. You basically wrote me out of the git blame record. I assume it was unintentional, but its not acceptable for me - when i do work I expect it to clearly appear I have done that.

sondrelg:
I see. Yep not intentional - I'll try to revert :)

sondrelg:
Done 👍 

image

Goldziher:
> Done 👍
> 
> image

much obliged.

PR #144, opened 2021-01-15
Remove djangorestframework camel case

Goldziher:


openapi_tester/configuration.py
```diff
@@ -20,11 +20,7 @@ def schema_loader(self):
 
     @property
     def case_tester(self) -> Callable:
-        return django_settings.OPENAPI_TESTER.get('CASE_TESTER', lambda *args: None)
-
-    @property
-    def camel_case_parser(self) -> bool:
-        return django_settings.OPENAPI_TESTER.get('CAMEL_CASE_PARSER', False)
+        return django_settings.OPENAPI_TESTER.get('CASE_TESTER', None)
```
sondrelg, line 23: Won't this reintroduce this issue?

https://github.com/snok/django-openapi-tester/issues/83#issuecomment-702828307

openapi_tester/schema_tester.py
```diff
@@ -79,79 +67,40 @@ def test_dict(self, schema: dict, data: dict, reference: str) -> None:
...
-                )
-                request_hint = 'You passed a None value where we expected a dict.'
-            raise DocumentationError(
-                message=f"Mismatched types. Expected item to be <class 'dict'> but found {type(data)}.",
-                response=data,
-                schema=schema,
-                reference=reference,
-                response_hint=response_hint or '',
-                request_hint=request_hint or '',
-            )
```
sondrelg, line 104: What will happen now if a response is documented as a dictionary, but contains some other type/is misdocumented?

Looks like it will fail a few lines down on line 113 `response_keys = data.keys()` instead.

openapi_tester/schema_tester.py
```diff
@@ -185,32 +133,6 @@ def test_list(self, schema: dict, data: Union[list, dict], reference: str) -> No
...
-                request_hint = 'You passed a None value where we expected a list.'
-            raise DocumentationError(
-                message=f"Mismatched types. Expected item to be <class 'list'> but found {type(data)}.",
-                response=data,
-                schema=schema,
-                reference=reference,
-                response_hint=response_hint or '',
-                request_hint=request_hint or '',
-            )
-
```
sondrelg, line 213: Same here

sondrelg (commented):
All djangorestframework_camel_case removals look good, and so a lot of the rewrites 👏  but what is the reason for removing the schema tester sections handling bad response types? Perhaps I'm missing something.

openapi_tester/schema_tester.py
Goldziher: nope, we are now using prance to validate the schema in its entirety. If the schema is invalid it will fail completely upon load. Otherwise we are raising an "expected" vs. "received" error, which is what the user needs.

openapi_tester/schema_tester.py
Goldziher: same as prev :)

openapi_tester/configuration.py
Goldziher: see my next comment.

Issue #147, opened 2021-01-15
Handle package renaming


sondrelg:
Creating this for the public record. 

Now that we're moving towards publishing a renamed package I think it's important to think about how we handle the renaming. My suggestion is to push updates to (and maintain) *both* versions on pypi going forward, forever. The reason I think this is necessary is that there just doesn't seem to be any support for package renaming on pypi in general, and this would prevent someone using automated dependency tooling from losing out on future versions of the package, if they've already defined `django-swagger-tester` as a dependency.

This should also be fairly easy to do; it should only be a matter of adding one or two extra steps to our existing package publishing pipeline. Concretely, I would image we could publish `django-openapi-tester` as before, then just `sed` the package name in the checked out `pyproject.toml` to `django-swagger-tester` and publish that. I also think maintaining a separate pypi-readme for the old package, where we can tell users to switch to the new package name, would be useful.

If things go to plan, I'll set this up this weekend.

sondrelg:
Decided against this, but we will need to push a new patch version to `django-swagger-tester` informing new users about the rename after we publish v1. of `django-openapi-tester`, so keeping the issue open until we do that.

sondrelg:
The latest and hopefully final naming change is from ´django-openapi-tester` to `drf-openapi-tester`, and has to do with which packages are our primary dependencies. After the refactor, DRF is primary dependency, so it seems fitting to name it that.

PR #148, opened 2021-01-16
Rework documentation structure

sondrelg:
Creates reusable rst-components we can include in our readthedocs files and readme's. 

We've previously maintained one root dir `README.rst` (now `README.md`), one `PYPI_README.rst`, and one `index.rst`, all with more or less the same content. This way we only have to write and update docs one place.


README.md
```diff
@@ -90,20 +92,6 @@ responses:
...
-          items:
-            type: object
-            properties:
-              id:
-                type: integer
-                format: int64
-              name:
-                type: string
-            xml:
-              name: Tag
```
sondrelg, line 106: I removed the `tags` section of the readme toy example. It's still really long, but was **really** long

docs/pypi/OLD_README.rst
```diff
@@ -0,0 +1,3 @@
+``django-swagger-tester`` has been renamed to ``django-openapi-tester``.
+
+If you're already using ``django-swagger-tester``, please update your requirements, and if you're looking for documentation, please see https://pypi.org/project/django-openapi-tester/ instead.
```
sondrelg, line 3: For the `django-swagger-tester` pypi readme. This relates to #147

sondrelg (commented):
Locally the readthedocs docs look the same before and after. ReadTheDocs also doesnt care where the docs are located, so think this should be pretty safe.

From their docs
```
Where do I need to put my docs for RTD to find it?

Read the Docs will crawl your project looking for a conf.py. 
Where it finds the conf.py, it will run sphinx-build in that directory. 
So as long as you only have one set of sphinx documentation in your project, it should Just Work.
```

JonasKs (approved):
Mmmmm, clean!

PR #149, opened 2021-01-16
Update pypi publishing workflow

sondrelg:
Updates ci workflow to publish a test-package before publishing to pypi


.github/workflows/publish_pypi.yml
```diff
@@ -7,21 +7,39 @@ on:
...
+      - uses: actions/setup-python@v2
         with:
-          python-version: 3.8
-      - name: Install poetry
+          python-version: 3.9
+      - uses: snok/install-poetry@v1.1.1
+
+      # django-openapi-tester was originally named django-swagger-tester.
+      # After renaming the package, we now maintain two packages, and so
+      # we publish to pypi twice.
```
Goldziher, line 20: why? i mean, no need to update the old package. Its going to be completely incompatible with the new code. I suggest simply deprecating the old package and publishing the new. 

We can start at v1.0 and give an explanation in the description of the package. Alas, we can keep the v3.0, but still give an explanation that there was a breaking change and the older version is deprecated.

sondrelg: After thinking about this a little, I agree. As long as existing users find their way back to the repo in some way we're fine. 

Instead of merging this, I will push a new patch version with no real changes to `django-swagger-tester` to update the pypi-readme so it informs readers of the renaming. Will do this after we release v1 of the new package.

JonasKs (approved):

PR #158, opened 2021-01-22
added new schemas and updated parsing logic

Goldziher:


openapi_tester/loaders.py
```diff
@@ -92,7 +92,7 @@ def de_reference_schema(self, schema: dict) -> dict:
             resolver.resolve_references()
             return resolver.specs
         except ResolutionError as e:
-            raise OpenAPISchemaError('infinite recursion error') from e
+            raise OpenAPISchemaError(e.args[0]) from e
```
sondrelg, line 95: What will this return?

openapi_tester/schema_converter.py
```diff
@@ -18,7 +18,9 @@ def __init__(self, schema: dict, with_faker: bool = False):
 
             Faker.seed(0)
             self.faker = Faker()
-        schema_type = schema['type']
+        schema_type = schema.get('type')
+        if not schema_type and 'properties' in schema:
+            schema_type = 'object'
```
sondrelg, line 23: Isn't there a case where `schema_type` is falsy, and 'properties' won't be in schema? Re the any type discussion

Edit: as in, should we handle the last case in addition to this if?

sondrelg (approved):
looks good 👏

PR #171, opened 2021-01-26
Pre-release cleanup

sondrelg:
Would recommend reading commit by commit. These are the final changes I want to make before pushing v0.1 to test. If you disagree with anything in the final commit, we can drop that.

### Changes

- Updates our metadata and package version
- Converts the changelog and contributing docs to markdown
- Applies some minor code suggestion refactors to simplify some conditionals
- Adds Django and DRF stubs 
- Lowers the accepted versions for our hard dependencies:
    - Inflection to ^0.4
    - Prance to ^0.16
    - Any for PyYaml


JonasKs (approved):

openapi_tester/schema_converter.py
```diff
@@ -8,7 +8,7 @@ class SchemaToPythonConverter:
     """ This class is used both by the DocumentationError format method and the various test suites """
 
     def __init__(self, schema: dict, with_faker: bool = False):
-        while any([keyword in schema for keyword in ["allOf", "oneOf", "anyOf"]]):
```
Goldziher, line 11: this is a regression: https://www.w3schools.com/python/ref_func_any.asp

openapi_tester/type_declarations.py
```diff
@@ -40,14 +40,20 @@ def export_anything(cls, module_name: str) -> None:
     the type checker does not actually run the code.
     """
     # noinspection PyUnresolvedReferences
-    from rest_framework.response import Response
+    from rest_framework.response import Response as RFResponse
```
Goldziher, line 43: I dont think this is good - instead, we should add the django rest framework mypy stubs. 

I can update our pre-commit to handle this.

Goldziher: to clarify, creating a special type for response is not good.

pyproject.toml
```diff
@@ -39,11 +42,11 @@ packages = [
 
 [tool.poetry.dependencies]
 djangorestframework = "*"
-inflection = "^0.5.1"
+inflection = "^0.4"
 openapi-spec-validator = "^0.2.9"
-prance = "^0.20.0"
-python = "^3.6.1"
-pyYAML = "^5.3.1"
+prance = "^0.16"
```
Goldziher, line 47: why the downgrade?

pyproject.toml
```diff
@@ -39,11 +42,11 @@ packages = [
 
 [tool.poetry.dependencies]
 djangorestframework = "*"
-inflection = "^0.5.1"
+inflection = "^0.4"
```
Goldziher, line 45: again, why the downgrade?

openapi_tester/schema_converter.py
sondrelg: How so? It should work exactly the same - the list comprehension is just redundant as any already accepts generators

openapi_tester/type_declarations.py
sondrelg: Ok stubs easily sounds like a better solution. I'll remove this.

pyproject.toml
sondrelg: It's not a downgrade, it's increasing the span we allow third parties to operate in. We're allowing
a user to use inflection in the range of `0.4 to  < 1.0` rather than `0.5.1 to < 1.0`. 

If we kept 0.5.1 as the lower bound and someone wanted to use this package, but were depending on `inflection==0.5.0`, Poetry would start raising resolver errors for them, when it's not strictly necessary.

pyproject.toml
sondrelg: Same here. Poetry will install the most recent version within a bound, so we're lowering the lower bound to give users more flexibility.

openapi_tester/type_declarations.py
sondrelg: Added in af458e2b7470a176db91aa2bf2c4275e6edae0d9

PR #176, opened 2021-01-27
Clean up workflows and handle mypy errors

sondrelg:
## Changes

- Removes sphinx from dev dependencies
- Cleans up workflows 
- Removes redundant mypy `type: ignore` comments
- Adds an exception when `_check_openapi_type` receives an unhandled type


openapi_tester/schema_tester.py
```diff
@@ -117,6 +118,7 @@ def _check_openapi_type(schema_type: str, value: Any, enum: Optional[List[Any]],
             return isinstance(value, dict)
         if schema_type == "array":
             return isinstance(value, list)
+        raise OpenAPISchemaError(f"Invalid schema type: {schema_type}")
```
sondrelg, line 121: This shouldn't ever happen, correct?

Goldziher: yes, no need to add this.

Goldziher: it shouldnt, no

PR #177, opened 2021-01-27
Add unit tests

sondrelg:
Adds some initial unit tests to make sure different combinations of inputs produce the desired results.


openapi_tester/schema_tester.py
```diff
@@ -90,17 +91,17 @@ def handle_one_of(
...
     @staticmethod
-    def _check_openapi_type(schema_type: str, value: Any, enum: Optional[List[Any]], format: Optional[str]) -> bool:
+    def _check_openapi_type(schema_type: str, value: Any, enum: Optional[List[Any]], _format: Optional[str]) -> bool:
         if enum:
             return value in enum
         if schema_type == "boolean":
             return isinstance(value, bool)
         if schema_type in ["string", "file"]:
-            if format == "bytes":
+            if _format == "byte":
```
sondrelg, line 100: Looks like the correct format is `byte` over `bytes` :) 

https://swagger.io/docs/specification/data-models/data-types/#string

Goldziher: why the variable name change?

openapi_tester/schema_converter.py
```diff
@@ -30,7 +32,8 @@ def __init__(self, schema: dict, with_faker: bool = False):
                 schema_type = "object"
             else:
                 raise ValueError(
-                    f"Schema type is not specified and cannot be inferred, please make sure to definte the type key for schema: {schema}"
+                    f"Schema type is not specified and cannot be inferred, "
+                    f"please make sure to definte the type key for schema: {schema}"
```
Goldziher, line 36: define*

openapi_tester/schema_converter.py
```diff
@@ -39,18 +42,18 @@ def __init__(self, schema: dict, with_faker: bool = False):
         else:
             self.result = self._to_mock_value(schema_type, schema.get("enum"), schema.get("format"))  # type :ignore
 
-    def _to_mock_value(self, schema_type: Any, enum: Optional[List[Any]], format: Optional[str]) -> Any:
+    def _to_mock_value(self, schema_type: Any, enum: Optional[List[Any]], _format: Optional[str]) -> Any:
```
Goldziher, line 45: why the name change to _format?

openapi_tester/schema_tester.py
```diff
@@ -152,11 +153,15 @@ def _route_error_text_addon(paths: KeysView) -> str:
...
-        return f'\n\nAvailable methods include: {", ".join(method.upper() for method in methods if method.upper() != "PARAMETERS")}.'
+        str_methods = ", ".join(method.upper() for method in methods if method.upper() != "PARAMETERS")
+        return f"\n\nAvailable methods include: {str_methods}."
 
     @staticmethod
     def _responses_error_text_addon(status_code: Union[int, str], response_status_codes: KeysView) -> str:
-        return f'\n\nDocumented responses include: {", ".join([str(key) for key in response_status_codes])}. Is the `{status_code}` response documented?'
+        return (
+            f'\n\nDocumented responses include: {", ".join([str(key) for key in response_status_codes])}. '
+            f"Is the `{status_code}` response documented?"
```
Goldziher, line 163: Perhaps a better phrasing:

Undocumented status-code `{status_code}`.\n\nDocumented responses include: ....

openapi_tester/schema_tester.py
```diff
@@ -292,15 +297,17 @@ def _test_openapi_type_object(
                     response=data,
                     schema=schema_section,
                     reference=reference,
-                    hint="You need to add the missing schema key to the response, or remove it from the documented response.",
+                    hint="You need to add the missing schema key to the response, "
```
Goldziher, line 300: I'd shorten this and all message, also remove any type of direct address to the user, i.e. "Add the missing key to the response or remove it from the schema."

openapi_tester/schema_tester.py
```diff
@@ -292,15 +297,17 @@ def _test_openapi_type_object(
...
+                    "or remove it from the documented response.",
                 )
             if response_key not in properties:
                 raise DocumentationError(
                     message=f"Key `{response_key}` not found in the OpenAPI schema.",
                     response=data,
                     schema=schema_section,
                     reference=reference,
-                    hint="You need to add the missing schema key to your documented response, or stop returning it in your API.",
+                    hint="You need to add the missing schema key to your documented "
```
Goldziher, line 309: same principle. Concise - if we cant fit it in one line, it should be shorter :)

openapi_tester/schema_tester.py
```diff
@@ -331,6 +338,7 @@ def _test_openapi_type_array(
                 hint="Document the contents of the empty dictionary to match the response object.",
             )
 
+        # noinspection PyTypeChecker
```
Goldziher, line 341: why? what is the inspection you are running?

openapi_tester/type_declarations.py
```diff
@@ -39,14 +37,10 @@ def export_anything(cls, module_name: str) -> None:
     We don't have to worry about circular imports here, since
     the type checker does not actually run the code.
     """
-    # noinspection PyUnresolvedReferences
-    from rest_framework.response import Response
-
-    # noinspection PyUnresolvedReferences
-    from rest_framework.test import APITestCase
+    from rest_framework.response import Response  # noqa: F401
```
Goldziher, line 40: this would be needed basically for every line. Its not good. You can disable the rule for the entire file, which is better.

I need to noinspection flags for intellij idea because it complains about unused imports. If you can turn this rule for the entire file, it would be good.

tests/test_schema_errors/test_arrays.py
```diff
@@ -0,0 +1,44 @@
...
+    tester.test_schema_section(example_schema_array, example_array, "")
+
+
+def test_empty_list():
+    """ An empty array should always pass """
+    tester.test_schema_section(example_schema_array, [], "")
+
+
+def test_null():
+    """ A null value should always raise an error """
```
Goldziher, line 22: it shouldnt, not if it has null: True, x-nullable etc.

tests/test_schema_errors/test_arrays.py
```diff
@@ -0,0 +1,44 @@
...
+    with pytest.raises(DocumentationError, match="Mismatched types, expected list but received NoneType"):
+        tester.test_schema_section(example_schema_array, None, reference="")
+
+
+def test_nullable():
+    """ A null value is allowed when the array is nullable """
+    schema = {"type": "array", "items": {"type": "string"}, "nullable": True}
+    tester.test_schema_section(schema, None, reference="")
```
Goldziher, line 30: reference should be made an optional param instead of being passed for every test. Just give it a default of "" in the method.

tests/test_schema_errors/test_arrays.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_wrong_type():
+    """ Type mismatches should raise errors """
+    with pytest.raises(DocumentationError, match="expected list but received dict."):
```
Goldziher, line 35: I wouldnt test the regex match, we might want to change these. Just check that "dict" is present. Otherwise, we can move the error messages to constants and THEN use them in the tests by importing the constants (constants.py is already in place), for example:

DOC_ERROR_MISMATCHED_TYPES_LIST_DICT =  "expected list but received dict."

tests/test_schema_errors/test_bad_type.py
```diff
@@ -0,0 +1,18 @@
...
+example_array = ["string"]
+
+tester = SchemaTester()
+
+
+def test_bad_schema_type():
+    """
+    Unsupported schema types should raise an error.
+    """
+    for _type in ["str", "list", "dict", "123", 123]:
```
Goldziher, line 15: Im not sure this test should be in place. This is testing a usage that CANNOT happen if the user uses the library as intended. We have a loader and validator in place that will not allow calling the test_schema_section method with this type of data. We therefore do not need to test it or handle this case in the code, and doing so (as was before) is over engineering that makes our lives harder.

tests/test_schema_errors/test_integers.py
```diff
@@ -0,0 +1,94 @@
+import pytest
```
Goldziher, line 1: this is a good test in general. Still, I would put this as part of the schema tester tests. You seem to displike large test files, but really below 500 lines a test file is ok in my book.

tests/test_schema_errors/test_integers.py
```diff
@@ -0,0 +1,94 @@
...
+    """ Not adhering to maximum limitations should raise an error """
+    with pytest.raises(
+        DocumentationError,
+        match="Mismatched content. Response integer violates the maximum value defined in the schema",
+    ):
+        tester.test_schema_section(example_schema_integer, 6, "")
+
+
+def test_null():
+    """ A null value should always raise an error """
```
Goldziher, line 63: this test is redundant and is not part of the test

tests/test_schema_errors/test_integers.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_nullable():
+    """ A null value is allowed when the integer is nullable """
+    schema = {"type": "integer", "nullable": True}
+    tester.test_schema_section(schema, None, reference="")
+
+
+def test_wrong_type():
+    """ Type mismatches should raise errors """
```
Goldziher, line 75: this test should occur only once, this is duplication.

tests/test_schema_errors/test_integers.py
```diff
@@ -0,0 +1,94 @@
...
+        tester.test_schema_section(example_schema_integer, 6, "")
+
+
+def test_null():
+    """ A null value should always raise an error """
+    with pytest.raises(DocumentationError, match="Mismatched types, expected int but received NoneType"):
+        tester.test_schema_section(example_schema_integer, None, reference="")
+
+
+def test_nullable():
```
Goldziher, line 68: same

tests/test_schema_errors/test_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+def test_maximum_violated():
+    """ Not adhering to maximum limitations should raise an error """
+    with pytest.raises(
+        DocumentationError, match="Mismatched content. Response number violates the maximum value defined in the schema"
+    ):
+        tester.test_schema_section(example_schema_number, 6, "")
+
+
+def test_null():
+    """ A null value should always raise an error """
```
Goldziher, line 65: duplication, please remove

tests/test_schema_errors/test_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+    with pytest.raises(DocumentationError, match="Mismatched types, expected int or float but received NoneType"):
+        tester.test_schema_section(example_schema_number, None, reference="")
+
+
+def test_nullable():
```
Goldziher, line 70: duplication, please remove

tests/test_schema_errors/test_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+    """ A null value is allowed when the number is nullable """
+    schema = {"type": "number", "nullable": True}
+    tester.test_schema_section(schema, None, reference="")
+
+
+def test_wrong_type():
+    """ Type mismatches should raise errors """
```
Goldziher, line 77: duplication, please remove

tests/test_schema_errors/test_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+    schema = {"multipleOf": 5, "type": "number"}
+    for number in [5, 10, 15, 20, 25]:
+        tester.test_schema_section(schema, number, "")
+
+    # Fail
+    with pytest.raises(DocumentationError, match="The response number must be a multiple of 5, but is 7."):
+        tester.test_schema_section(schema, 7, "")
+
+
+# endregion: number
```
Goldziher, line 97: ?

tests/test_schema_errors/test_objects.py
```diff
@@ -0,0 +1,52 @@
...
+tester = SchemaTester()
+
+
+def test_nothing_wrong():
+    """ This should always pass """
+    tester.test_schema_section(example_schema_object, example_object, "")
+
+
+def test_response_is_missing_keys():
+    """ Missing keys in a response object should raise an error """
```
Goldziher, line 17: nope, this test is too crude. You should read the logic and test in a more subtle and elaborate way.

tests/test_schema_errors/test_objects.py
```diff
@@ -0,0 +1,52 @@
...
+    with pytest.raises(DocumentationError, match="The following properties are missing from the tested data: value"):
+        tester.test_schema_section(example_schema_object, {}, "")
+
+
+def test_schema_object_is_missing_keys():
+    """ Excess keys in a response should raise an error """
```
Goldziher, line 23: same regarding all tests in this file.

tests/test_schema_errors/test_objects.py
```diff
@@ -0,0 +1,52 @@
...
+    with pytest.raises(DocumentationError):
+        schema = {"type": "object", "properties": {}}
+        tester.test_schema_section(schema, example_object, "")
+
+
+def test_null():
+    """ A null value should always raise an error """
+    with pytest.raises(DocumentationError, match="Mismatched types, expected dict but received NoneType"):
```
Goldziher, line 31: duplication, please remove

tests/test_schema_errors/test_objects.py
```diff
@@ -0,0 +1,52 @@
...
+        tester.test_schema_section(example_schema_object, None, "")
+
+
+def test_nullable():
+    """ A null value is allowed when the object is nullable """
```
Goldziher, line 36: duplication, please remove

tests/test_schema_errors/test_objects.py
```diff
@@ -0,0 +1,52 @@
...
+    schema = {"type": "object", "properties": {"value": {"type": "integer"}}, "nullable": True}
+    tester.test_schema_section(schema, None, "")
+
+
+def test_wrong_type():
+    """ Type mismatches should raise errors """
+    with pytest.raises(DocumentationError, match="expected dict but received list."):
```
Goldziher, line 43: duplication, please remove

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+
+from openapi_tester import DocumentationError, SchemaTester
+
+example_schema_string = {"type": "string", "minLength": 3, "maxLength": 5}
+example_string = "str"
+
+tester = SchemaTester()
+
+
+# region: string
```
Goldziher, line 11: whats the point of these comments?

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+def test_max_length_violated():
+    """ Not adhering to maxlength limitations should raise an error """
+    with pytest.raises(
+        DocumentationError,
+        match="Mismatched content. Response string violates the maximum string " "length defined in the schema",
+    ):
+        tester.test_schema_section(example_schema_string, "a" * 6, "")
+
+
+def test_null():
```
Goldziher, line 37: duplication, please remove

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+    """ A null value should always raise an error """
+    with pytest.raises(DocumentationError, match="Mismatched types, expected str but received NoneType"):
+        tester.test_schema_section(example_schema_string, None, reference="")
+
+
+def test_nullable():
```
Goldziher, line 43: duplication, please remove

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+    """ A null value is allowed when the array is nullable """
+    schema = {"type": "string", "nullable": True}
+    tester.test_schema_section(schema, None, reference="")
+
+
+def test_wrong_type():
```
Goldziher, line 49: duplication, please remove

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+
+# region: special formats
+
+
+def test_good_date():
+    tester.test_schema_section({"type": "string", "format": "date"}, "2040-01-01", "")
+
+
+def test_bad_date():
+    with pytest.raises(DocumentationError, match="expected a value with the format date but received 01-31-2019"):
```
Goldziher, line 73: we can handle a lot of different date formats. This wont cut it. Put a really invalid date value.

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+
+def test_good_datetime():
+    """ Should pass """
+    tester.test_schema_section({"type": "string", "format": "date-time"}, "2040-01-01 08:00", "")
+
+
+def test_bad_datetime():
+    with pytest.raises(
+        DocumentationError, match="expected a value with the format date-time but received 2040-01-01 0800"
+    ):
```
Goldziher, line 85: same, whatever django can parse we can parse as well because we use the django parsers.

tests/test_schema_errors/test_strings.py
```diff
@@ -0,0 +1,110 @@
...
+    with pytest.raises(DocumentationError, match="expected a value with the format byte but received test"):
+        tester.test_schema_section({"type": "string", "format": "byte"}, "test", "")
+
+
+# endregion: special formats
+
+
+def test_pattern():
+    """ The a regex pattern can be passed to describe how a string should look """
+    schema = {"type": "string", "pattern": r"^\d{3}-\d{2}-\d{4}$"}
```
Goldziher, line 103: this is not going to be simple

openapi_tester/schema_tester.py
sondrelg: just so we're not overwriting the `format` builtin. 

It shouldn't matter for the code in this repo, but I think there's a risk that we can change behavior in downstream code by overwriting it.

https://www.programiz.com/python-programming/methods/built-in/format

openapi_tester/type_declarations.py
sondrelg: So the noinspection flags aren't necessary for me when there's a `noqa` flag, so I thought it would be better to have `n` noqa comments than `2n + 1`  lines of code to manage the noqa file comment + all the noinspect lines.

If the inline `noqa` doesn't work for you, lets revert :)

tests/test_schema_errors/test_arrays.py
sondrelg: Yep, but we're explicitly passing a schema element where it's not nullable. I originally named the test, `test_null_when_not_nullable`, but the next test is already called `test_nullable`, so I shortened it.

openapi_tester/schema_converter.py
sondrelg: See reply in other question

openapi_tester/schema_tester.py
sondrelg: Changed!

openapi_tester/schema_tester.py
sondrelg: What do you think about changing this to `It is missing from the response or vice versa.`?

I was considering removing the hint altogether, but it does actually seem useful.

openapi_tester/schema_tester.py
sondrelg: I wonder if this is just a problem with my setup on my work laptop. The Pycharm built in type hinting complains that data is `None` here, but it doesn't seem to be a problem on my other computer (and I'm guessing not yours). I'll remove it.

openapi_tester/type_declarations.py
sondrelg: I see the same as you on my home comp, so reverting 👍

tests/test_schema_errors/test_arrays.py
sondrelg: I'll consolidate the two tests to this, so it's more clear

```python
def test_nullable():
    # A null value should always raise an error
    with pytest.raises(DocumentationError, match="Mismatched types, expected list but received NoneType"):
        tester.test_schema_section(example_schema_array, None, reference="")
    # Unless the schema specifies it should be nullable
    schema = {"type": "array", "items": {"type": "string"}, "nullable": True}
    tester.test_schema_section(schema, None, reference="")
```

tests/test_schema_errors/test_arrays.py
sondrelg: Done

tests/test_schema_errors/test_bad_type.py
sondrelg: Yeah I don't disagree with this. I'll remove the test.

tests/test_schema_errors/test_integers.py
sondrelg: I structured this mostly, to easily progress through the data types docs and put tests in place for the different format validation, etc. that we want to support. 

Now that they're most in place though, I don't mind consolidating them. There's a lot of repeated tests, like the nullable test for every data type, that probably should be one test iterating through all the data types. 

I'll apply other suggestions, then I'll consolidate tests into one file in a separate commit 👍

tests/test_schema_errors/test_numbers.py
sondrelg: image

tests/test_schema_errors/test_objects.py
sondrelg: How is it too crude? This seems completely fundamental to me, if we're attempting any sort of validation. If an API is documented to return something, and it isn't returned,  we should raise an exception.

Are you agreeing with the above, but suggesting this isn't a good way to test this, because of how we've structured the new validation?

tests/test_schema_errors/test_strings.py
sondrelg: They tell your IDE to create collapsible sections of code. Makes it easier to separate sections.

image

tests/test_schema_errors/test_strings.py
sondrelg: According to this https://swagger.io/docs/specification/data-models/data-types/#string we should **only** support [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) 

I purposefully put in a valid datetime format that just doesn't match ISO8601.

tests/test_schema_errors/test_strings.py
sondrelg: See response above. This seems like it goes against the OAS spec - but I might be wrong.

tests/test_schema_errors/test_strings.py
sondrelg: hehe, is it not just a matter of trying to create a regex  and testing it? 

```python
import re


try:
    compiled_pattern = re.compile(pattern)
except re.error:
    raise one error
if not compiled_pattern.match(string):
    raise another error
```

sondrelg:
```python
import re
  ...: 
  ...: pattern = r"^\d{3}-\d{2}-\d{4}$"
  ...: string = "test"
  ...: 
  ...: try:
  ...:     compiled_pattern = re.compile(pattern)
  ...: except re.error:
  ...:     print('bad pattern')
  ...: if not compiled_pattern.match(string):
  ...:     print('bad string')
  ...:     
>> bad string
import re
  ...: 
  ...: pattern = r"^\d{3}-\d{2}-\d{4}$"
  ...: string = "123-12-1234"
  ...: 
  ...: try:
  ...:     compiled_pattern = re.compile(pattern)
  ...: except re.error:
  ...:     print('bad pattern')
  ...: if not compiled_pattern.match(string):
  ...:     print('bad string')
  ...:     
>>
```

openapi_tester/schema_tester.py
sondrelg: I guess it might not matter since we're in the scope of the single function, but also don't see a very good reason to use builtin namespaces if we can avoid it

Goldziher: ok, in general i prefer to use _ for private class methods and private variables, for function scope variables, just informative naming. So this could be schema_format or format_type etc. I also do not think we have a problem with the builtin spacename - its a safe use case.

openapi_tester/schema_tester.py
Goldziher: i think its less clear. we should be as simple and straightforward as possible, i.e. this should be something you can brainlessly understand :)

openapi_tester/type_declarations.py
Goldziher: so this is up to intellij inspection settings. Some of the insepections are global, and i dont want to turn these off. The imports one seems to be like this (tell me if you find a better way.) --> p.s. we can also have a .idea folder to share settings, if its helpful.

tests/test_schema_errors/test_arrays.py
Goldziher: yup, is better IMO.

tests/test_schema_errors/test_numbers.py
Goldziher: still dont get it.

tests/test_schema_errors/test_objects.py
Goldziher: yes, i agree we should test this. But the test is crude vis-a-vis the validation logic. 

We are using the required keyword to decide what is required and what isn't. Only if a key is specified in the required list, or if no required list is supplied (we then assume everything is required) do we regard it as missing.

tests/test_schema_errors/test_strings.py
Goldziher: i see, interesting. Thanks!

tests/test_schema_errors/test_strings.py
Goldziher: Fair enough, BUT - this is a question we should consider: are we trying to follow the openapi specification to the letter? I think we should add a configurable "strict" flag, which dictates this behaviour. As a user of the library - i might decide to serialize/deserialize dates in different formats for the consumption of FE clients (e.g. a react based website for example). In that case what matters is that the date or datetime is a valid date / datetime string, within the Django eco-system. This is why I am using the django parsers for validation here.

tests/test_schema_errors/test_strings.py
Goldziher: you are right, but we need to decide on this. See prev.

tests/test_schema_errors/test_strings.py
Goldziher: why not :), go ahead?

tests/test_schema_errors/test_objects.py
sondrelg: Nice, change it to this

```python
def test_response_is_missing_keys():
    # If a required key is missing, we should raise an error
    required_key = {"type": "object", "properties": {"value": {"type": "integer"}}, 'required': ['value']}
    with pytest.raises(DocumentationError, match="The following properties are missing from the tested data: value"):
        tester.test_schema_section(required_key, {})

    # If not required, it should pass
    optional_key = {"type": "object", "properties": {"value": {"type": "integer"}}}
    tester.test_schema_section(optional_key, {})
```

tests/test_schema_errors/test_strings.py
sondrelg: Implemented in bcea60c 👍

tests/test_schema_errors/test_strings.py
sondrelg: A strict flag sounds good. Perhaps we can add a separate issue for this, and handle it outside this PR?

openapi_tester/schema_tester.py
sondrelg: Do you have a suggestion for what we could use?

How about `The response should contain this key or the documentation should change`?

tests/test_schema_errors/test_strings.py
sondrelg: And move into the class in db3df73

PR #183, opened 2021-01-31
Split validation into methods

sondrelg:
@Goldziher, this is how I pictured it. 

This shouldn't (unless I've made a mistake) change the code in any way, except for structure, and I think it makes it quite a bit clearer. What do you think?


Goldziher (approved):
Approved

PR #184, opened 2021-01-31
Add validators

sondrelg:
Adds validation logic to make our existing tests pass, and fixes a few bugs found along the way

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -255,18 +255,35 @@ class SchemaTester:
 
     @staticmethod
     def _validate_multiple_of(schema_section: dict, data: Union[int, float]) -> Union[Optional[str], bool]:
-        str_multiple = schema_section.get("multipleOf")
+        multiple = schema_section.get("multipleOf")
 
-        if not str_multiple:
+        if multiple is None:
             return False
 
-        try:
-            multiple_of = int(str_multiple) if int(str_multiple) == float(str_multiple) else float(str_multiple)
-        except ValueError:
-            raise OpenAPISchemaError("Unable to parse multipleOf as a number")
+        if not data % multiple == 0:
+            return f"The response value {data} should be a multiple of {multiple}"
 
-        if not data % multiple_of == 0:
-            return f"The response value {data} should be a multiple of {multiple_of}"
+    @staticmethod
+    def _validate_min_and_max(schema_section: dict, data: Union[int, float]) -> Union[Optional[str], bool]:
+        minimum = schema_section.get("minimum")
+        if minimum is not None:
+            exclusive_minimum = schema_section.get("exclusiveMinimum")
+            if not exclusive_minimum:
+                if data < minimum:
+                    return f"The response value {data} exceeds the minimum allowed value of {minimum}"
+            else:
+                if data <= minimum:
+                    return f"The response value {data} exceeds the minimum allowed value of {minimum + 1}"
+
+        maximum = schema_section.get("maximum")
+        if maximum is not None:
+            exclusive_maximum = schema_section.get("exclusiveMaximum")
+            if not exclusive_maximum:
+                if data > maximum:
+                    return f"The response value {data} exceeds the maximum allowed value of {maximum}"
+            else:
+                if data >= maximum:
+                    return f"The response value {data} exceeds the maximum allowed value of {maximum - 1}"
 
     def test_schema_section(
         self,
@@ -304,6 +321,7 @@ class SchemaTester:
                 self._validate_pattern,
                 self._validate_enum,
                 self._validate_multiple_of,
+                self._validate_min_and_max,
             ]
             for validator in validators:
                 error = validator(schema_section, data)
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -236,7 +236,9 @@ class SchemaTester:
 
     def _validate_openapi_type(self, schema_section: dict, data: Any) -> Optional[str]:
         valid = True
-        schema_type: str = schema_section["type"]
+        schema_type = schema_section.get("type")
+        if not schema_type:
+            return None
         if schema_type in ["string", "file"]:
             valid = isinstance(data, (str, bytes))
         elif schema_type == "integer":
@@ -313,7 +315,7 @@ class SchemaTester:
         schema_section_type = schema_section.get("type")
         if not schema_section_type and "properties" in schema_section:
             schema_section_type = "object"
-        if schema_section_type is None or data is None and self.is_nullable(schema_section):
+        if schema_section_type is None or (data is None and self.is_nullable(schema_section)):
             # If there`s no type, any response is permitted so we return early
             # If data is None and nullable, we also return early
             return
@@ -348,6 +350,7 @@ class SchemaTester:
                 self._validate_min_and_max,
                 self._validate_length,
             ]
+
             for validator in validators:
                 error = validator(schema_section, data)
                 if isinstance(error, str):
@@ -388,15 +391,17 @@ class SchemaTester:
         message, hint = "", ""
         for required_key in required_keys:
             if required_key not in response_keys:
-                missing_keys = ", ".join(str(key) for key in sorted(list(set(required_keys) - set(response_keys))))
                 hint = "Remove the key(s) from your OpenAPI docs, or include it in your API response."
-                message = f"The following properties are missing from the tested data: {missing_keys}."
+                message = f"The following property is missing from the tested data: {required_key}."
             else:
                 response_keys.remove(required_key)
         for response_key in response_keys:
             if response_key not in properties:
                 hint = "Remove the key(s) from your OpenAPI docs, or include it in your API response."
-                message = f"The following properties was found in the response, but is missing from the schema definition: {response_key}."
+                message = (
+                    f"The following property was found in the response, "
+                    f"but is missing from the schema definition: {response_key}."
+                )
 
         if message:
             raise DocumentationError(
@@ -407,10 +412,10 @@ class SchemaTester:
                 hint=hint,
             )
 
-        for schema_key, response_key in zip([key for key in properties.keys() if key in data.keys()], data.keys()):
+        for schema_key, response_key in zip([key for key in properties if key in data], data):
             self._validate_key_casing(schema_key, case_tester, ignore_case)
             self._validate_key_casing(response_key, case_tester, ignore_case)
-            if schema_key in required_keys and schema_key not in data.keys():
+            if schema_key in required_keys and schema_key not in data:
                 raise DocumentationError(
                     message=f"Schema key `{schema_key}` was not found in the tested data.",
                     response=data,
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -185,11 +185,11 @@ class SchemaTester:
             tester(key)
 
     @staticmethod
-    def _validate_enum(schema_section: dict, data: Any) -> Union[Optional[str], bool]:
+    def _validate_enum(schema_section: dict, data: Any) -> Optional[str]:
         enum = schema_section.get("enum")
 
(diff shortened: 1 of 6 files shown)


openapi_tester/schema_tester.py
```diff
@@ -185,20 +185,24 @@ def _validate_key_casing(
             tester(key)
 
     @staticmethod
-    def _validate_enum(schema_section: dict, data: str) -> Union[Optional[str], bool]:
-        if "enum" not in schema_section:
+    def _validate_enum(schema_section: dict, data: Any) -> Union[Optional[str], bool]:
+        enum = schema_section.get("enum")
+
+        if enum is None:
```
Goldziher, line 191: if not enum rather than using is.

Goldziher: ''return not enum or data in enum''

Also, you can put as default value to get an empty list.

openapi_tester/schema_tester.py
```diff
@@ -185,20 +185,24 @@ def _validate_key_casing(
...
                 f'Mismatched values, expected a member of the enum {schema_section["enum"]} but received {str(data)}.'
             )
 
     @staticmethod
-    def _validate_pattern(schema_section: dict, data: str) -> Union[Optional[str], bool]:
-        if "pattern" not in schema_section:
+    def _validate_pattern(schema_section: dict, data: Any) -> Union[Optional[str], bool]:
+        pattern = schema_section.get("pattern")
+
+        if pattern is None:
```
Goldziher, line 203: same principle for all.

openapi_tester/schema_tester.py
```diff
@@ -185,20 +185,24 @@ def _validate_key_casing(
...
-        if enum and data not in enum:
+        if data not in enum:
             return (
                 f'Mismatched values, expected a member of the enum {schema_section["enum"]} but received {str(data)}.'
             )
 
     @staticmethod
-    def _validate_pattern(schema_section: dict, data: str) -> Union[Optional[str], bool]:
-        if "pattern" not in schema_section:
+    def _validate_pattern(schema_section: dict, data: Any) -> Union[Optional[str], bool]:
```
Goldziher, line 200: Why return a union type from validators? This shoukd be consisteny - optional[str]

openapi_tester/schema_tester.py
```diff
@@ -264,7 +312,23 @@ def test_schema_section(
...
         This method orchestrates the testing of a schema section
         """
-        if "oneOf" in schema_section and data is not None:
+        schema_section_type = schema_section.get("type")
+        if not schema_section_type and "properties" in schema_section:
+            schema_section_type = "object"
+        if schema_section_type is None or data is None and self.is_nullable(schema_section):
+            # If there`s no type, any response is permitted so we return early
+            # If data is None and nullable, we also return early
+            return
```
sondrelg, line 321: Need to handle bad inputs before moving to `oneOf` logic

openapi_tester/schema_tester.py
```diff
@@ -253,6 +257,50 @@ def _validate_openapi_type(self, schema_section: dict, data: str) -> Union[Optio
                 f"but received {type(data).__name__}."
             )
 
+    @staticmethod
+    def _validate_multiple_of(schema_section: dict, data: Any) -> Union[Optional[str], bool]:
+        multiple = schema_section.get("multipleOf")
+
+        if multiple is None:
+            return False
```
Goldziher, line 265: return or return None

openapi_tester/schema_tester.py
```diff
@@ -319,19 +380,26 @@ def _test_openapi_type_object(
...
-            properties = {}
-        required_keys = schema_section["required"] if "required" in schema_section else list(properties.keys())
-        response_keys = data.keys()
-
-        if len([key for key in response_keys if key in required_keys]) != len(required_keys):
-            missing_keys = ", ".join(str(key) for key in sorted(list(set(required_keys) - set(response_keys))))
-            hint = "Remove the key(s) from your OpenAPI docs, or include it in your API response."
-            message = f"The following properties are missing from the tested data: {missing_keys}."
+
+        required_keys = schema_section.get("required", [])
```
sondrelg, line 387: Note: The default used to be `list(properties.keys())`, but I've changed it to `{}` so all keys are treated as optional unless specified

openapi_tester/schema_tester.py
```diff
@@ -253,6 +257,50 @@ def _validate_openapi_type(self, schema_section: dict, data: str) -> Union[Optio
...
+                if data <= minimum:
+                    return f"The response value {data} exceeds the minimum allowed value of {minimum + 1}"
+
+        maximum = schema_section.get("maximum")
+        if maximum is not None:
+            exclusive_maximum = schema_section.get("exclusiveMaximum")
+            if not exclusive_maximum:
+                if data > maximum:
+                    return f"The response value {data} exceeds the maximum allowed value of {maximum}"
+            else:
```
Goldziher, line 288: elif

openapi_tester/schema_tester.py
```diff
@@ -253,6 +257,50 @@ def _validate_openapi_type(self, schema_section: dict, data: str) -> Union[Optio
...
+                    return f"The response value {data} exceeds the minimum allowed value of {minimum}"
+            else:
+                if data <= minimum:
+                    return f"The response value {data} exceeds the minimum allowed value of {minimum + 1}"
+
+        maximum = schema_section.get("maximum")
+        if maximum is not None:
+            exclusive_maximum = schema_section.get("exclusiveMaximum")
+            if not exclusive_maximum:
+                if data > maximum:
```
Goldziher, line 286: can be combined with the containing if clause. 

I suggest adding pylint for this.

openapi_tester/schema_tester.py
```diff
@@ -253,6 +257,50 @@ def _validate_openapi_type(self, schema_section: dict, data: str) -> Union[Optio
...
+                    return f"The response value {data} exceeds the maximum allowed value of {maximum}"
+            else:
+                if data >= maximum:
+                    return f"The response value {data} exceeds the maximum allowed value of {maximum - 1}"
+
+    @staticmethod
+    def _validate_length(schema_section: dict, data: str) -> Union[Optional[str], bool]:
+        min_length = schema_section.get("minLength")
+        if min_length is not None:
+            if len(data) < min_length:
```
Goldziher, line 296: Again the if clauses can be combined

openapi_tester/schema_tester.py
```diff
@@ -253,6 +257,50 @@ def _validate_openapi_type(self, schema_section: dict, data: str) -> Union[Optio
...
+                return f"The length of {data} exceeds the minimum allowed length of {min_length}"
+
+        max_length = schema_section.get("maxLength")
+        if max_length is not None:
+            if len(data) > max_length:
```
Goldziher, line 301: Same

openapi_tester/schema_tester.py
```diff
@@ -264,7 +312,23 @@ def test_schema_section(
...
         """
-        if "oneOf" in schema_section and data is not None:
+        schema_section_type = schema_section.get("type")
+        if not schema_section_type and "properties" in schema_section:
+            schema_section_type = "object"
+        if schema_section_type is None or data is None and self.is_nullable(schema_section):
+            # If there`s no type, any response is permitted so we return early
+            # If data is None and nullable, we also return early
+            return
+        if data is None:
```
Goldziher, line 322: not data

openapi_tester/schema_tester.py
```diff
@@ -319,19 +380,26 @@ def _test_openapi_type_object(
...
+        for required_key in required_keys:
+            if required_key not in response_keys:
+                missing_keys = ", ".join(str(key) for key in sorted(list(set(required_keys) - set(response_keys))))
+                hint = "Remove the key(s) from your OpenAPI docs, or include it in your API response."
+                message = f"The following properties are missing from the tested data: {missing_keys}."
+            response_keys.remove(required_key)
+        for response_key in response_keys:
+            if response_key not in properties:
+                hint = "Remove the key(s) from your OpenAPI docs, or include it in your API response."
+                message = f"The following properties was found in the response, but is missing from the schema definition: {response_key}."
```
sondrelg, line 400: Note that we're removing values as we iterate through the list, so we're left with only non-required keys. Out of this reduced list we can then verify that they are at least documented, and otherwise throw an error.

tests/test_test_schema_section.py
```diff
@@ -78,21 +78,19 @@ def test_wrong_type():
 
 
 def test_min_length_violated():
-    # TODO: Make this pass (not implemented yet)
     """ Not adhering to minlength limitations should raise an error """
     with pytest.raises(
         DocumentationError,
-        match="Mismatched content. Response string violates the minimum string " "length defined in the schema",
+        match="The length of aa exceeds the minimum allowed length of 3",
```
Goldziher, line 84: aa?

tests/test_test_schema_section.py
```diff
@@ -78,21 +78,19 @@ def test_wrong_type():
...
         tester.test_schema_section(example_schema_string, "a" * 2)
 
 
 def test_max_length_violated():
-    # TODO: Make this pass (not implemented yet)
     """ Not adhering to maxlength limitations should raise an error """
     with pytest.raises(
         DocumentationError,
-        match="Mismatched content. Response string violates the maximum string " "length defined in the schema",
+        match="The length of aaaaaa exceeds the maximum allowed length of 5",
```
Goldziher, line 93: You are testing specific messages. This is an antipattern. 

I suggest moving the error messages into constants and use .format:

MY_MSG = "message {0}"

And then use the constant in both the code and test.

Goldziher (approved):
Please see my comments.

openapi_tester/schema_tester.py
sondrelg: Agree - done in e111565

openapi_tester/schema_tester.py
sondrelg: Updated 👍 453a53e

openapi_tester/schema_tester.py
sondrelg: updated

openapi_tester/schema_tester.py
sondrelg: Yeah perhaps - this seems easier to read though, no? This way the comparisons are aligned

openapi_tester/schema_tester.py
sondrelg:
```python
maximum = schema_section.get("maximum")
if maximum is not None:
    exclusive_maximum = schema_section.get("exclusiveMaximum")
    if not exclusive_maximum and data > maximum:
            return f"The response value {data} exceeds the maximum allowed value of {maximum}"
    elif data >= maximum:
            return f"The response value {data} exceeds the maximum allowed value of {maximum - 1}"
```

This just seems much easier to read than this

```python
maximum = schema_section.get("maximum")
if maximum is not None:
    exclusive_maximum = schema_section.get("exclusiveMaximum")
    if not exclusive_maximum:
        if data > maximum:
            return f"The response value {data} exceeds the maximum allowed value of {maximum}"
    else:
        if data >= maximum:
            return f"The response value {data} exceeds the maximum allowed value of {maximum - 1}"
```
but say the word and I'll change 👍

openapi_tester/schema_tester.py
sondrelg: Don't think this is right. An empty string, array, or object should have their schemas assessed - here we're checking nullability, so we're actually just after `None`s

tests/test_test_schema_section.py
sondrelg: It's just the printed value of the string passed as `data`:

```python
tester.test_schema_section(example_schema_string, "a" * 2)
```

Maybe we should add quotes to the print.

tests/test_test_schema_section.py
sondrelg: Yep, agree. Think I will do that in a separate PR following this one.

PR #188, opened 2021-02-06
Add anyOf handling

sondrelg:
## Changes

- Splits the allOf handler into three methods -> `handle_all_of`, `_combine_schemas`, and `_merge_dict` for reusability and easier testing
- Changed the flow of the `allOf` handler to match the flow of the `anyOf` and `oneOf` handler - this makes the base method cleaner IMO
- Updates some of our `reference` variables. We might need to do a full check to see that we're logging this appropriately. The problem here was that we weren't adding anything to the reference variable as we traversed the `anyOf`, etc. handlers
- Added a method for generating combinations of schemas, `get_all_combinations`
- Added `anyOf` handling that brute-force checks all combinations for matches before raising an error
- Added tests to exactly replicate the official docs on `anyOf`

And a final change that we can revert or improve: added some debug loggers to help debug the flow.

TODO: 

- [ ] Correct SchemaConverter to handle new allOf flow.

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -9,6 +9,7 @@ from rest_framework.test import APITestCase
 
 from openapi_tester import type_declarations as td
 from openapi_tester.constants import (
+    ANY_OF_ERROR,
     EXCESS_RESPONSE_KEY_ERROR,
     INVALID_PATTERN_ERROR,
     MISSING_RESPONSE_KEY_ERROR,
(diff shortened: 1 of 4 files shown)


openapi_tester/schema_tester.py
```diff
@@ -61,17 +66,43 @@ def __init__(
             raise ImproperlyConfigured("No loader is configured.")
 
     @staticmethod
-    def handle_all_of(**kwargs: dict) -> dict:
-        properties: Dict[str, Any] = {}
-        for entry in kwargs.pop("allOf"):
```
sondrelg, line 66: Using pop mutates our test examples, so I've replaced them with normal indexes

PR #189, opened 2021-02-06
feat(pylint)

Goldziher:
This PR adds pylint configs, and fixes pylint issues.

Please note- I disabled quite a few pylint rules. Some because I find them weird, others because we have failures that I do not have the time to fix. I suggest someone (@sondrelg ?) removes these ignores at some point and runs "pre-commit run --all-files" to see the errors. Maybe fix some of them, or not.


sondrelg (approved):
LGTM 👏

.github/workflows/testing.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
-      - run: python -m pip install pre-commit
+      - run: python -m pip install pre-commit pylint
```
sondrelg, line 22: Don't think this should be required. Pre-commit should handle all the package installs on its own, like it does with flake8, black, etc.

Was it failing?

Issue #187, opened 2021-02-02
Missing OAS features


sondrelg:
Now that we're getting closer to a stable release, I think now might be a good time to create a complete list of the remaining features we want should implement.

The schema combination stuff is by far the most complex and most important, but once that's done, it looks like we could add the remaining validation in a few hours.

Here are all the different features I could find ~in the order described on [swagger.io](https://swagger.io/docs/specification/data-models/):

## Features we've added, or will add shortly

- [x] allOf
- [x] anyOf
- [x] oneOf 
- [ ] ~not~ (we will not add this, as we believe it is already implicitly handled by existing logic)
- [x] [type](https://swagger.io/docs/specification/data-models/data-types/) validation
- [x] format validation
    - [x] byte
    - [x] date
    - [x] date-time
    - [x] double
    - [x] float
    - [ ] binary
    - [ ] ~password~ (out of scope)
- [x] nullable validation
- [x] enum validation 
- [x] support for [any](https://swagger.io/docs/specification/data-models/data-types/#any) type
- [x] minimum and maximum, including exclusive minimum and maximum validation
- [x] multiple of validation
- [x] min and max length validation
- [x] pattern validation
- [x] required property validation

## Other
- [x] unique items validation
- [x] array length validation
- [x] read-only and write-only property validation
- [x] additional properties/free-form objects handling
- [x] validating number of properties

Feel free to add to the list if anything is missing.

Goldziher:
im not sure regarding password and binary. Binary is possible, password is difficult to do and redundant as well.

sondrelg:
Agreed, password seems out of scope, and to some degree binary does too.

Goldziher:
I added writeOnly logic - at least partially.

PR #191, opened 2021-02-07
Validators

sondrelg:
Adds three missing validators as documented in #187 

- Minimum and maximum array lengths
- Minimum and maximum number of properties lengths
- Unique items


sondrelg:
@Goldziher, the new validators cause some of the larger schema tests to fail, because we are generating data that is not valid. If you know how to fix this, please feel free to commit a fix directly to the branch 👍

Goldziher (approved):

PR #196, opened 2021-02-07
added additional properties logic

Goldziher:


sondrelg (approved):
For the sake of posterity, can you quickly explain in the summary what we're doing to handle `additionalProperties` and in what ways?

PR #199, opened 2021-02-07
Add tests

sondrelg:
# Changes

- Adds a few tests
- Moves all schema tester tests into the same file, and splits them by class
- Adds a `show_expected` kwarg to our `DocumentationError` to choose whether we should show the entire formatted error, or just the error message. Just the error message makes more sense (IMO) when we're raising anyOf, oneOf, and some of the validation errors. Maybe disabling it for all validation errors is a mistake.
- Updates the missing loader error message
- Removes try/except block for date parser, since it will never raise an error, and is therefore untestable
- Adds boolean type validation (which we've somehow missed 🤷‍♂️ )


openapi_tester/exceptions.py
```diff
@@ -8,26 +8,30 @@ class DocumentationError(AssertionError):
...
 
-        converted_schema = SchemaToPythonConverter(schema, with_faker=False).result
-        super().__init__(
-            self.format(
-                response=self._sort_data(response),
-                example_item=self._sort_data(converted_schema),
-                hint=hint,
-                message=message,
-                reference=reference,
+        if show_expected:
```
Goldziher, line 22: this is not a good solution IMO. 

Two options:

1. create a root DRFTesterException and then sub-class it. 
2. make the parameters optional in this exception, and based on (a) if we have a schema passed, a response passed etc, run format or do not.
3. I personally think that option 1 is better - more pythonic/standard on the one hand, and less complex. The errors in this library are by far the most verbose I have seen :) and the format method here is overengineered IMHO. I.E. i'd strip this to the essential.

openapi_tester/schema_tester.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         elif "drf_yasg" in settings.INSTALLED_APPS:
             self.loader = DrfYasgSchemaLoader()
         else:
-            raise ImproperlyConfigured("No loader is configured.")
+            raise ImproperlyConfigured("Unable to infer which loader to use.")
```
Goldziher, line 72: Why is that a better message? it says the same in more words, some of which require better english :).

openapi_tester/schema_tester.py
```diff
@@ -267,31 +269,33 @@ def _validate_format(schema_section: dict, data: Any) -> Optional[str]:
             valid = isinstance(data, bytes)
         elif schema_format in ["date", "date-time"]:
             parser = parse_date if schema_format == "date" else parse_datetime
-            try:
-                result = parser(data)
-                valid = result is not None
-            except ValueError:
-                valid = False
+            result = parser(data)
```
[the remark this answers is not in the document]
Goldziher: valid = parser(data) is not None

openapi_tester/schema_tester.py
```diff
@@ -267,31 +269,33 @@ def _validate_format(schema_section: dict, data: Any) -> Optional[str]:
...
+                schema_type = "object"
+            elif "items" in schema_section:
+                schema_type = "array"
+            else:
+                return None
         if schema_type in ["file", "string"]:
             valid = isinstance(data, (str, bytes))
         elif schema_type == "integer":
-            valid = isinstance(data, int)
+            valid = isinstance(data, int) and not isinstance(data, bool)
```
Goldziher, line 290: why?

openapi_tester/schema_tester.py
```diff
@@ -267,31 +269,33 @@ def _validate_format(schema_section: dict, data: Any) -> Optional[str]:
...
         elif schema_type == "number":
-            valid = isinstance(data, (int, float))
+            valid = isinstance(data, (int, float)) and not isinstance(data, bool)
```
Goldziher, line 292: why?

Goldziher (changes requested):

openapi_tester/schema_tester.py
sondrelg: Only because the user cannot "configure" a loader, so it felt hard to act on. 

We can change it to something else, but it's generally nice that the error message hints to the root cause of the error, and I think the current message is a little outdated. 

Do you have any suggestions? We could use `Unable to configure loader` if you prefer that

openapi_tester/schema_tester.py
sondrelg: Because the `bool` class in Python inherits from integer. We therefore need to be careful about how we validate OAS types. 

```python
>>> isinstance(True, int)
True
```

The spec very clearly states that *only* `True`/`False` values are valid `boolean` OAS types (not truthy values like `1` to represent `True`), and so I would imagine `True`/`False` also should not be considered valid `number` or `integer` types.

In other words, this is a difference between the native python types behavior and the spec, and we therefore need this in place to make sure we're validating the types correctly. Without this, we would validate `True` with an integer schema as correct.

https://swagger.io/docs/specification/data-models/data-types/#boolean

openapi_tester/schema_tester.py
sondrelg: see answer above :)

openapi_tester/exceptions.py
sondrelg: Yeah I don't disagree with this. A base exception sounds like a good solution; I'll give it another try tonight.

openapi_tester/exceptions.py
```diff
@@ -8,26 +8,30 @@ class DocumentationError(AssertionError):
...
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         message: str,
         response: Any,
         schema: dict,
         hint: str = "",
         reference: str = "",
+        show_expected=True,
```
sondrelg, line 18:
```suggestion
        show_expected: bool = True,
```

openapi_tester/schema_tester.py
```diff
@@ -267,31 +269,33 @@ def _validate_format(schema_section: dict, data: Any) -> Optional[str]:
             valid = isinstance(data, bytes)
         elif schema_format in ["date", "date-time"]:
             parser = parse_date if schema_format == "date" else parse_datetime
-            try:
-                result = parser(data)
-                valid = result is not None
-            except ValueError:
-                valid = False
+            result = parser(data)
+            valid = result is not None
```
sondrelg, line 273:
```suggestion
            valid = parser(data) is not None
```

sondrelg:
lets squash this PR when it's good btw

openapi_tester/schema_tester.py
Goldziher: but the solution for this is not to add the redundant checks here, its to change the order:

```
>>> isinstance(False, bool)
True
>>> isinstance(1, bool)
False
```

you see?

Goldziher: put the bool test above the int test and thats that

openapi_tester/schema_tester.py
Goldziher: same as what i wrote above as well

openapi_tester/exceptions.py
Goldziher: well, please fix this in the scope of this PR - not a separate one. I do not want this "show_expected" stuff to be merged at all.

openapi_tester/exceptions.py
```diff
@@ -8,33 +8,40 @@ class DocumentationError(AssertionError):
...
+            super().__init__(message)
 
     @staticmethod
     def _sort_data(data_object: Any) -> Any:
         if isinstance(data_object, dict):
             return dict(sorted(data_object.items()))
         if isinstance(data_object, list):
-            return sorted(data_object)
+            try:
+                return sorted(data_object)
```
Goldziher, line 42: this is redundant - we already check for isinstance above, that why i removed the try catch block before.

openapi_tester/loaders.py
```diff
@@ -167,6 +169,69 @@ def resolve_path(self, endpoint_path: str) -> tuple:
             )
         raise ValueError(f"Could not resolve path `{endpoint_path}`")
 
+    @staticmethod
```
Goldziher, line 172: this logic should not be moved to the loaders, it should remain part of the tester and that the only place where its used. Please stop refactoring stuff just for the sake of refactoring it.

openapi_tester/loaders.py
```diff
@@ -167,6 +169,69 @@ def resolve_path(self, endpoint_path: str) -> tuple:
...
+        """
+        try:
+            return schema[key]
+        except KeyError as e:
+            raise UndocumentedSchemaSectionError(
+                UNDOCUMENTED_SCHEMA_SECTION_ERROR.format(key=key, error_addon=error_addon)
+            ) from e
+
+    @staticmethod
+    def _get_status_code(schema: dict, status_code: Union[str, int], error_addon: str = "") -> dict:
```
Goldziher, line 185: same

openapi_tester/loaders.py
```diff
@@ -167,6 +169,69 @@ def resolve_path(self, endpoint_path: str) -> tuple:
...
+        if str(status_code) in schema:
+            return schema[str(status_code)]
+        if int(status_code) in schema:
+            return schema[int(status_code)]
+        raise UndocumentedSchemaSectionError(
+            UNDOCUMENTED_SCHEMA_SECTION_ERROR.format(key=status_code, error_addon=error_addon)
+        )
+
+    def get_response_schema_section(self, response: td.Response) -> dict:
+        """
```
Goldziher, line 198: same

openapi_tester/schema_tester.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         elif "drf_yasg" in settings.INSTALLED_APPS:
             self.loader = DrfYasgSchemaLoader()
         else:
```
Goldziher, line 71: you can remove the else here.

openapi_tester/schema_tester.py
```diff
@@ -148,71 +149,9 @@ def handle_any_of(
             schema=schema_section,
             reference=f"{reference}.anyOf",
             hint="",
+            show_expected=False,
         )
 
-    @staticmethod
```
Goldziher, line 153: please revert this entire change.

openapi_tester/exceptions.py
sondrelg: Ok no problem.

openapi_tester/schema_tester.py
sondrelg: Hmm, well this isn't broken logic; in fact it seems like isolated, each of these lines are how we should validate the OAS type - and your proposition will mean any reordering will break our validation in the future.

```python
>>> isinstance(False, int) and not isinstance(False, bool)
False
>>> isinstance(1, int) and not isinstance(1, bool)
True
```

But sure, I can reorder them and remove the extra type check

openapi_tester/exceptions.py
sondrelg: The removal broke the logic (and a test). Sorting can break with `TypeError` depending on the contents of the list.

openapi_tester/schema_tester.py
sondrelg: This actually won't work. Even if we move the block up, this isn't the filtering element; we filter by oas type:

```python
elif schema_type == "boolean":
    valid = isinstance(data, bool)
elif schema_type == "integer":
    valid = isinstance(data, int) and not isinstance(data, bool)
elif schema_type == "number":
    valid = isinstance(data, (int, float)) and not isinstance(data, bool)
```

sondrelg: The boolean block will never hit to eliminate the `isinstance(data, bool)` scenario. The whole point here is that we should prevent an `"integer"` type from being validated successfully if it's a python boolean, because even though that's fine in python, it's not in the OAS spec.

openapi_tester/exceptions.py
sondrelg:
```python
sorted([1,2,4,3])
Out[2]: [1, 2, 3, 4]
sorted([{},{}])
Traceback (most recent call last):
  File "/Users/sondrelg/Documents/django-openapi-tester/.venv/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3343, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-96c4907b4da2>", line 1, in <module>
    sorted([{},{}])
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

sondrelg: In this case we want to default to returning them unsorted, which is what we achieve with the try/except block

openapi_tester/schema_tester.py
sondrelg: done

sondrelg:
BTW @Goldziher, I've just removed the exception stuff for now. We can fix it in a separate PR. For now, I want to fix the boolean validation.

openapi_tester/exceptions.py
Goldziher: Cool

openapi_tester/schema_tester.py
Goldziher: Ok

Goldziher (approved):

PR #200, opened 2021-02-08
cleanup

Goldziher:


openapi_tester/schema_tester.py
```diff
@@ -513,16 +513,6 @@ def test_openapi_array(
...
-        if data is None:
-            if self.is_nullable(schema_section):
-                return
-            raise DocumentationError(
-                message=NONE_ERROR.format(expected="list"),
-                response=data,
-                schema=schema_section,
-                reference=f"{reference}.array",
-                hint="Document the contents of the empty dictionary to match the response object.",
-            )
```
sondrelg, line 525: Are we handling this somewhere else?

sondrelg (approved):
Nice 👏 Just one question

openapi_tester/schema_tester.py
Goldziher: Yup, but we should do this just once,not twice

PR #201, opened 2021-02-09
Schema validation unit

sondrelg:


openapi_tester/exceptions.py
```diff
@@ -65,15 +73,15 @@ def __init__(self, key: str, case: str, expected: str) -> None:
         super().__init__(f"The response key `{key}` is not properly {case}. Expected value: {expected}")
 
 
-class OpenAPISchemaError(Exception):
+class LoadingError(Exception):
```
Goldziher, line 76: this name is worse then the previous name - LoadingError is too generic. OpenAPISchemaError is a better name.

openapi_tester/loaders.py
```diff
@@ -167,6 +169,69 @@ def resolve_path(self, endpoint_path: str) -> tuple:
             )
         raise ValueError(f"Could not resolve path `{endpoint_path}`")
 
+    @staticmethod
```
Goldziher, line 172: this logic doesnt belong in the loader. the loader just loads the schema.

sondrelg: Just saw this now. I also made this change in the #199 PR before I saw this. 

Not sure I 100% get this separation though. The method calls the loader internals and *loads* the response schema section. For the purpose of this package it seems like loading the schema is redundant - the response schema is the unit we're after, and so it feels fine to put in the loaders.py imo.

Nothing in `schema_tester.py` touches this method except `validate_response` which also has access to `self.loader`

sondrelg:
Closing for now

PR #202, opened 2021-02-09
Update error formatting

sondrelg:
This 

```
---------------------------------------------------------------------------
DocumentationError                        Traceback (most recent call last)
<ipython-input-2-f8a914c30544> in <module>
----> 1 raise DocumentationError('Error message', response="1", schema={'type': 'integer'})

DocumentationError: Error: Mismatched values, expected a value with the format int but received str

Expected: "int"

Received: "1"
```

Becomes this

```
---------------------------------------------------------------------------
DocumentationError                        Traceback (most recent call last)
<ipython-input-2-f8a914c30544> in <module>
----> 1 raise DocumentationError('Error message', response="1", schema={'type': 'integer'})

DocumentationError: Mismatched values, expected a value with the format int but received str

Expected: 42

Received: "1"
```

### DIFF
--- a/openapi_tester/constants.py
+++ b/openapi_tester/constants.py
@@ -10,6 +10,23 @@ OPENAPI_PYTHON_MAPPING = {
     "integer": int.__name__,
     "number": f"{int.__name__} or {float.__name__}",
 }
+OPENAPI_TYPE_EXAMPLES = {
+    "boolean": True,
+    "string": "string",
+    "file": "image.jpeg",
+    "array": [],
+    "object": {},
+    "integer": 42,
+    "number": 4.2,
+}
+OPENAPI_FORMAT_EXAMPLES = {
+    "date": "2020-10-01",
+    "date-time": "2020-10-01 20:00:00",
+    "double": 2.2,
+    "float": 2.2,
+    "byte": b"bytestring",
+}
+
 PARAMETER_CAPTURE_REGEX = re.compile(r"({[\w]+})")
 
 # Validation errors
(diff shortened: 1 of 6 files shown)


Goldziher:
But why?.

I intentionally placed the type there 😃.

openapi_tester/constants.py
```diff
@@ -10,6 +10,23 @@
     "integer": int.__name__,
     "number": f"{int.__name__} or {float.__name__}",
 }
+OPENAPI_TYPE_EXAMPLES = {
```
Goldziher, line 13: i intentionally used typings - i thiknk its clearer and more precise. So this change I am against (both constants here.). 

The rest of the PR is good. So please revert this particular change and then lets merge this.

sondrelg: Sure can do. Just one follow up question: what do you feel about changing `json.dumps(example_item)` to just `example_item`? 

The confusing thing (to me) as that we're wrapping `int` in `"int"`. If we said something like 

```
Expected type: int

Received: "1"
```

I think it would make it much less confusing, potentially

sondrelg: Checked, and while it makes the `int` example clearer, it probably makes more complex structures more confusing, as it just replaces all the double quotes with single quotes.

````
{"key1": "str", "key2": "str", "key3": "str", "key4": ["int"]}

sondrelg: Reverted to the old map

sondrelg: If we straight up replace quotes, we end up with `{key1: str, key2: str, key3: str, key4: [int]}`. Think that's an improvement.

Goldziher (approved):

PR #210, opened 2021-02-15
moved coveragerc into pyproject.toml

Goldziher:


sondrelg:
While we're at it, lets also move the `pytest.ini`
https://docs.pytest.org/en/stable/customize.html

sondrelg (approved):

PR #214, opened 2021-02-17
feat(added new format validators

Goldziher:


openapi_tester/validators.py
```diff
@@ -25,6 +28,37 @@
...
+            return False
+
+    return wrapped
+
+
+_number_format_validator = _create_format_validator(
+    lambda x: isinstance(x, float) if x != 0 else isinstance(x, (int, float)), True
+)
+
+DJANGO_VALIDATOR_MAP = {
```
sondrelg, line 46: Small detail: not sure about the naming; why did you call it *django* validator map?

sondrelg (commented):
Think this looks good so far 👏  Not sure what the best thing would be for testing this - there's no need to test DRF code, but it would be nice to have unit tests for each validator to proof against future changes to the code.

openapi_tester/validators.py
```diff
@@ -28,37 +28,58 @@
...
+    schema_type: str = schema_section.get("type", "object")
+    schema_format: str = schema_section.get("format", "")
+    if schema_format in DJANGO_VALIDATOR_MAP and not DJANGO_VALIDATOR_MAP[schema_format](data):
+        return VALIDATE_FORMAT_ERROR.format(expected=schema_section["format"], received=str(data))
+    if not DJANGO_VALIDATOR_MAP[schema_type](data):
+        return VALIDATE_TYPE_ERROR.format(
+            expected=OPENAPI_PYTHON_MAPPING[schema_type],
+            received=type(data).__name__,
+        )
+    return None
```
sondrelg, line 80: I don't see a good reason for combining these validators. The checks here are not dependent on each other as far as I can tell, so we should keep them in different functions.

I suspect you might have consolidated them because all the validators are in a shared map; but If anything I think it would make more sense to have a separate type and format validator map, to not blur the lines.

Goldziher: No, there needs to be conditional logic for format validation. Specifically for the ''byte'' format, which is type string in OA, buy bytes in python. 

The alternative is cruder, allow bytes for type ''string'',and i dobt like it.

sondrelg: Could we not just run the format validator before the type validator and keep them separate?

openapi_tester/schema_tester.py
```diff
@@ -308,9 +308,10 @@ def test_schema_section(
             return
 
         schema_section_type = self.get_schema_type(schema_section)
+        if not schema_section_type:
+            return
         validators: List[Callable] = [
-            validate_enum,
-            validate_openapi_type,
+            validate_type,
```
sondrelg, line 314: should we put `validate_format` first?

sondrelg (approved):
Looks great 👍 

Think we should write more tests before publishing a new version, but for now this seems fine to merge 🎉

PR #217, opened 2021-02-20
Failing test for issue 203

Goldziher:


Goldziher:
Still need to update tests and the readme

openapi_tester/loaders.py
```diff
@@ -128,34 +130,54 @@ def resolve_path(self, endpoint_path: str) -> tuple:
...
             message += f"\n\nDid you mean one of these?{closest_matches}"
         raise ValueError(message)
 
+    @staticmethod
+    def handle_pk_parameter(resolved_route: ResolverMatch, path: str, method: str) -> Tuple[str, ResolverMatch]:
+        split_view_path = resolved_route.view_name.split(".")
+        view_name = split_view_path.pop()
+        imported_module = import_module(".".join(split_view_path))
+        view = getattr(imported_module, view_name)
+        coerced_path = BaseSchemaGenerator().coerce_path(path=path, method=method, view=view)
```
sondrelg, line 165:
```suggestion
        model = getattr(resolved_route.func.view_class, 'model', None)
        field_name = get_pk_name(model) if model else 'id'
```
Drawing on the `coerce_path` method logic I think we might be able to shorten it to ~this. Unless there are cases where the `view_name` is included in a resolved parth, but `view_class` isn't; but that seems unlikely

openapi_tester/loaders.py
```diff
@@ -128,34 +130,54 @@ def resolve_path(self, endpoint_path: str) -> tuple:
         parsed_path = url_object.path if url_object.path.endswith("/") else url_object.path + "/"
         if not parsed_path.startswith("/"):
             parsed_path = "/" + parsed_path
+        for key, value in self.field_key_map.items():
```
sondrelg, line 133: So this enables us to optionally override, for things like the i18n feature, while coerced DRF paths should be automatically replaced?

sondrelg (approved):
Looks good 👏👏👏

PR #223, opened 2021-02-25
Replace functools cached_property wrapper with django wrapper

sondrelg:


Goldziher:
lots of test failures, please make sure these are only our indeterminant test 😉

Goldziher (approved):

Issue #224, opened 2021-02-25
Nullable enums not validating


idesoto-rover:
Hi, I'm having the following issue. `drf-spectacular` for some reason generates nullable enums like a `oneOf` of two enum types, one which contains the enum options and a NullEnum containing only the None option. So for example, if I have the following field in my model:
```
    building_type = models.CharField(
        max_length=25,
        choices=(('hs', 'House'), ('apt', 'Apartment'), ('farm', 'Farm')),
        default=None,
        null=True,
    )
```
It will generate the following field in my schema:
```
...
        building_type:
          nullable: true
          oneOf:
          - $ref: '#/components/schemas/BuildingTypeEnum'
          - $ref: '#/components/schemas/NullEnum'
...
    BuildingTypeEnum:
      enum:
      - hs
      - apt
      - farm
      type: string
    NullEnum:
        enum:
        - null
```

So the issue I'm having is that `drf-openapi-tester` fails to validate this when using a valid enum option (e.g. `hs` in the example). This is because `handle_one_of` method will fail if the value matches both of the types. It's matching the `BuildingTypeEnum` correctly because `hs` is one of the options in the enum, but it's also matching the NullEnum because of these two lines in `test_schema_section`:
```
        schema_section_type = self.get_schema_type(schema_section)
        if not schema_section_type:
            return
```
Since NullEnum doesn't have a `type` or a `properties` field, `get_schema_type` is returning None and so it will exit early without running the validators.

The quickest fix I can think of to support this case is to change those two lines to:
```
        schema_section_type = self.get_schema_type(schema_section)
        if not schema_section_type and 'enum' not in schema_section:
            return
```
but maybe theres' some more generic fix?

FTR, this is what drf-spectacular says about this NullEnum type: https://github.com/tfranzel/drf-spectacular/issues/235

Goldziher:
Hi, thanks for your report.

I believe the behaviour of DRF spectacular is in fact wrong. Use of oneOf and the null enum are a proposal, not a part of the spec.

I will see about extending support for this. For the time being though, you can turn off the the DRF spectacular behaviour in that library's configs.

idesoto-rover:
Thanks! Yes, that was my plan B, to disable the NullEnum in drf-spectacular. But when I do that it will generate something like this:
```
        building_type:
          allOf:
          - $ref: '#/components/schemas/BuildingTypeEnum'
          nullable: true
```
[...]

PR #227, opened 2021-02-27
fix/224: updated schema testing and conversion

Goldziher:
This PR addresses #224. 

- added a field to the Names model that follows the example in the issue
- updated the drf spectacular schemas
- updated the test utils to ensure this is properly tested
- updated the schema converter to handle "oneOf" used for enums
- updated the schema tester to handle  "oneOf" used for enums
- resolved some typing issues in the loader class (unrelated - but mypy complained)


openapi_tester/loaders.py
```diff
@@ -62,9 +65,10 @@ def get_schema(self) -> dict:
         """
         Returns OpenAPI schema.
         """
-        if self.schema is None:
-            self.set_schema(self.load_schema())
-        return self.schema  # type: ignore
+        if self.schema:
+            return self.schema
+        self.set_schema(self.load_schema())
+        return self.get_schema()
```
sondrelg, line 71: mypy rewrite?

sondrelg (approved):
Looks solid. Don't have much to add, so go ahead and merge/publish when you want 👏

openapi_tester/loaders.py
Goldziher: yup

PR #229, opened 2021-02-27
fix/208: updated handling of anyOf

Goldziher:


sondrelg (approved):
Looks great 🚀  It's going to be so nice to have passing tests again 🎉

PR #230, opened 2021-02-28
updated pre-commit

Goldziher:


sondrelg (approved):

.pre-commit-config.yaml
```diff
@@ -12,21 +12,20 @@ repos:
       - id: trailing-whitespace
       - id: check-ast
       - id: check-json
+      - id: check-yaml
       - id: check-merge-conflict
-      - id: no-commit-to-branch
```
Goldziher, line 16: one doesnt have privileges or one knows what he is doing :)

.pre-commit-config.yaml
```diff
@@ -12,21 +12,20 @@ repos:
...
-        args: [ --branch, master ]
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.4
     hooks:
       - id: flake8
         additional_dependencies: [
-            'flake8-bugbear==20.11.1',
```
Goldziher, line 23: removed pinned - i assume the latest version works for us

sondrelg: Yep agree

.pre-commit-config.yaml
sondrelg: Just fyi, you can always skip pre-commit by running `--no-verify` when you commit, so not sure this needs to be removed

Goldziher: im aware but i do not like -n

PR #239, opened 2021-10-20
Avoid KeyError when utilizing additionalProperties=true

alexpaulzor:


sondrelg:
Hi @alexpaulzor!

Can you explain a little bit about the motivation for this change? What's not behaving as expected, and do you have some code to illustrate?

alexpaulzor:
Ok @sondrelg I was finally able to get unit tests running locally and I added a simple test that fails without my change, but works with it. Thank you, and I'm sorry it took so long.

sondrelg:
Thanks a lot @alexpaulzor, this looks good to me! 👏 

I'll release a new version right away 🙂

sondrelg:
Released v1.3.10 now with the fix 👍

PR #243, opened 2021-11-02
Allow empty response content

### LINKED ISSUE #238: Cannot validate response with empty content
Hello!
I'm really excited about this project because it perfectly fit my current needs - I want to have a custom open api file, but verify if it's not out of sync with the API.

However, there's one problem I'm having now I need your help.
When I test response of an endpoint that returns no content, I get this error:
```
E           openapi_tester.exceptions.UndocumentedSchemaSectionError: Error: Unsuccessfully tried to index the OpenAPI schema by `content`. 
E           
E           No content documented for method: delete, path: /endpoint/{version}/path
```

I've looked it up in code and it is raised in the `SchemaTester.get_key_value`, which raises error when a key is not defined.
It is however possible to have no `content` key. This way you define an empty response in OpenAPI 3.x.

How could I get it fixed? Should I skip testing empty responses?
I think I could provide a fix for this, but please tell me what is the desired way of solving this problem.

radowit:
Closes #238

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -140,17 +140,28 @@ class SchemaTester:
         if "openapi" not in schema:  # pylint: disable=E1135
             # openapi 2.0, i.e. "swagger" has a different structure than openapi 3.0 status sub-schemas
             return self.get_key_value(status_code_object, "schema")
-        content_object = self.get_key_value(
-            status_code_object,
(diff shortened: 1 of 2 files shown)


sondrelg:
Nice work @radowit 👍 This looks good to me, so happy to accept it as is 🙂 Thanks!

I'll release a new version now.

Issue #245, opened 2021-12-16
Incorrect validation for additionalProperties?


idesoto-rover:
Hello!

According to https://swagger.io/docs/specification/data-models/dictionaries/, you can specify a dictionary that contains string values like this:
```
type: object
additionalProperties:
  type: string
```

If the dictionary can contain any type of value, you would specify it as either of these two options:
```
type: object
additionalProperties: true
```

```
type: object
additionalProperties: {}
```

Currently, `drf-openapi-tester` only considers valid the `additionalProperties: true` option because of [this line of code](https://github.com/snok/drf-openapi-tester/blob/master/openapi_tester/schema_tester.py#L311) that only considers extra keys valid if `additionalProperties` is `true`.

When `additionalProperties` is a dictionary, the code is checking whether the keys are defined inside `additionalProperties`, but I don't think this is correct. I think the correct validation would be: if `additionalProperties` is a dictionary, validate that the values for any keys that are not defined in the `properties` section match the schema defined in additionalProperties. So for example, in the first example above we would just be validating that values are of type `string`. If the schema for `additionalProperties` is of type `object`, then we validate that the value matches that object specification (the value itself contains the keys specified in the `properties` section of the `additionalProperties` schema).

Let me try to show an example using an `additionalProperties` of type `object`:
For this schema:
```
type: object
properties:
    key_1:
        type: string
additionalProperties:
    type: object
    properties:
        key_2:
            type: string
        key_3:
            type: string
```
This data should be valid:
```
{
    'key_1': 'value_1',
    'some_extra_key': {
        'key_2': 'value_2',
        'key_3': 'value_3'
    }
}
```
But not this one because `key_2` and `key_3` are supposed to belong to the values of any extra keys, but not to the object itself. The validation error here should be that we were expecting an `object` for keys that are not `key_1` but we got a `string`:
```
{
    'key_1': 'value_1',
    'key_2': 'value_2',
    'key_3': 'value_3'
}
```
And the error here is that it's missing `key_1` which is specified in the schema:
```
{
    'some_extra_key': {
        'key_2': 'value_2',
        'key_3': 'value_3'
    }
}
```

I could try to fix this myself, but I would like confirmation that I'm not misunderstanding something and that what I describe should be the expected behaviour.

Thanks!

Goldziher:
Hi there. Thanks for the detailed analysis. I think you are correct. We would very much appreciate a PR 😉

sondrelg:
[...]

PR #246, opened 2021-12-17
#245 fix validation for additionalProperties

idesoto-rover:
Fixes issues described in #245:
- allows specifying a schema in `additionalProperties` and validates any extra keys in the response that are not defined in the `properties` section against this schema
- `additionalProperties` can be an empty schema `{}` and this means any values are allowed in extra keys
- `additionalProperties` can be `True` and this behaves the same as an empty schema (this was already working)

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -295,8 +295,7 @@ class SchemaTester:
         required_keys = [key for key in schema_section.get("required", []) if key not in write_only_properties]
         response_keys = data.keys()
         additional_properties: Optional[Union[bool, dict]] = schema_section.get("additionalProperties")
-        if not properties and isinstance(additional_properties, dict):
-            properties = additional_properties
+        additional_properties_allowed = additional_properties is not None
         for key in properties.keys():
             self.test_key_casing(key, case_tester, ignore_case)
             if key in required_keys and key not in response_keys:
@@ -307,9 +306,7 @@ class SchemaTester:
                 )
         for key in response_keys:
             self.test_key_casing(key, case_tester, ignore_case)
-            key_in_additional_properties = isinstance(additional_properties, dict) and key in additional_properties
-            additional_properties_allowed = additional_properties is True
-            if key not in properties and not key_in_additional_properties and not additional_properties_allowed:
+            if key not in properties and not additional_properties_allowed:
                 raise DocumentationError(
                     f"{VALIDATE_EXCESS_RESPONSE_KEY_ERROR.format(excess_key=key)}\n\nReference: {reference}.object:key:"
                     f"{key}\n\nHint: Remove the key from your API response, or include it in your OpenAPI docs"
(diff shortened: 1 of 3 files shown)


Goldziher:
Nice!

Do we need an update to the readme as well?

@sondrelg can you add sonar scanner so coverage is picked?

sondrelg:
Looks like the tests are failing on Python 3.10. From what I can tell it's not related to the PR, we just need to add [this line](https://github.com/snok/asgi-correlation-id/blob/main/pyproject.toml#L66) to our dev dependencies in master.

idesoto-rover:
> Do we need an update to the readme as well?

I took a look and didn't find anything in the README file that would be affected by these changes. What did you have in mind?

I thought I could add a line to the `CONTRIBUTING.md` file explaining how to run tests, though. I wasn't familiar with `poetry` and it took me a while to figure it out. Would that be OK?

sondrelg:
Go for it 👏

Goldziher:
LGTM. Thanks!

sondrelg:
I can look over this later tonight and release a patch, unless you're already doing so @Goldziher 👍

Goldziher:
> I can look over this later tonight and release a patch, unless you're already doing so @Goldziher 👍 

Nope, go ahead 😃

openapi_tester/schema_tester.py
```diff
@@ -295,8 +295,7 @@ def test_openapi_object(
         required_keys = [key for key in schema_section.get("required", []) if key not in write_only_properties]
         response_keys = data.keys()
         additional_properties: Optional[Union[bool, dict]] = schema_section.get("additionalProperties")
-        if not properties and isinstance(additional_properties, dict):
-            properties = additional_properties
+        additional_properties_allowed = additional_properties is not None
```
sondrelg, line 298: Since we specify 

```python
elif isinstance(additional_properties, dict):
```

we currently won't have any handling for non-bool/non-dict types. What do you think about amending this assignment to this, so all data types are handled?

```suggestion
        additional_properties_allowed = additional_properties is True or isinstance(additional_properties, dict)
```

sondrelg (commented):
Looks good, just have one question!

openapi_tester/schema_tester.py
idesoto-rover: `additionalProperties` can only be a boolean or an object, so I think those two would be roughly equivalent. The only difference, if the schema is not considered invalid by the schema validation, would be in how we want to handle schemas where `additionalProperties` is defined as a different type.

So say there's a schema defined like this:
```
type: object
additionalProperties: 1
```

Do we want to allow `additionalProperties` in this case or not? With the `is not None` check we would allow them but not validate anything about them. With your proposed change we would not allow `additionalProperties`. I don't mind either option.

sondrelg: From what I can gather, it would be possible for users to specify whatever they want, but anything other than a dict/bool/null would be a violation of the spec - does that seem right? If that's correct, we're deciding whether we think it's within the scope of this library to raise an error for the invalidity of the schema or not.

I can also see both sides, but am probably leaning towards raising an error here. Perhaps instead of the suggestion I made above, we should do something like this?

```python
if not isinstance(additional_properties, (bool, dict, None)):
    raise OpenAPISchemaError('Invalid additional properties type')
additional_properties_allowed = additional_properties is not None
```

What do you think @Goldziher? 🙂

Goldziher: yes, i agree @sondrelg - i think your last suggestion is the cleanest of what has been suggested so far.

idesoto-rover: I've added the check and a test. Changed it a little bit because `None` cannot be used as a type in `isinstance`.

Goldziher:
@idesoto-rover - thanks for your work.

Linting is failing though. 

I will let @sondrelg handle it from here.

sondrelg:
Nice 👏 Thanks @idesoto-rover!

I'll fix the linting when bumping the version 👍

PR #249, opened 2021-12-19
Add sonarcloud coverage upload

sondrelg:
PR adds sonarcloud uploads w/ coverage reporting. To do so we changed the `--cov` source from `openapi_tester` to `.` which means we need to ignore a few tests to get back to 100% test coverage.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -81,11 +81,32 @@ jobs:
       - name: Run tests
         run: |
           source .venv/bin/activate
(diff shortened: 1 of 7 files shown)


sondrelg:
@Goldziher any idea why sonarcloud is missing coverage information here? Turned of auto-scan and can't see a problem with the properties file.

Goldziher:
Is the coverage file in the correct place? Also check the output of the sonar cli

sondrelg:
Yeah looks to be an issue with paths relative to source directory. I'll work it out.

Issue #248, opened 2021-12-18
v2 release


sondrelg:
Opening this to document ~~breaking~~ all changes we want to include for a v2 release shortly.

I think we should:

- Add an `__all__` module export in our `openapi/__init__.py` (see https://github.com/snok/drf-openapi-tester/pull/247/commits/3de9ee2191eebf63336bf431ed825d5b3ee4908f), where we also remove some of our exports. I don't think exporting the `OPENAPI_PYTHON_MAPPING` constant makes sense - it's an internal resource.
- Drop Python 3.6 and upgrade relevant language features 
- Add Django 4.0 support to pyproject.toml and our Github matrix. It looks like Django 4.0 breaks two tests for some mysterious reason - we'll need to figure out why and resolve the issue. As a part of this change we should upgrade our demo project where needed.

If anyone has suggestions for other changes, feel free to document them here.

PR #250, opened 2021-12-19
Drop Python 3.6

sondrelg:
Relates to #248. 

PR includes backwards incompatible changes meant for the v2 release; mainly:

- We drop support for Python 3.6 (though currently just in the metadata - it will still work)
- We trim the module exports to only expose public functions and classes. I don't think constants and exceptions should be exported, as they're not intended to be used downstream.


setup.cfg
```diff
@@ -59,3 +48,21 @@ exclude =
     manage.py,
     .venv
 max-complexity = 16
+
+[mypy]
```
Goldziher, line 52: you can move this to pyproject.toml

Goldziher (approved):

setup.cfg
sondrelg: good point, I'll do that in a follow-up PR

PR #253, opened 2021-12-19
Update supported Django versions

sondrelg:
Relates to #248.

PR drops Django 2.2 and adds Django 4.0. The intention is to be left with Django 3.0+, and Python 3.7+ support.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ exclude =
 max-complexity = 16
 
 [mypy]
-python_version = 3.9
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -51,7 +51,7 @@ exclude =
 max-complexity = 16
 
 [mypy]
-python_version = 3.9
+python_version = 3.10
```
Goldziher, line 54: you dont need to specify a version at all.

PR #258, opened 2022-04-09
Allow empty null response

radowit:
Empty response worked only if `response.json` returned something.
But if a response data is not json-serializable schema tester raises an error.

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -161,7 +161,7 @@ class SchemaTester:
             )
             return self.get_key_value(json_object, "schema")
 
-        if response.json():
+        if response.data and response.json():
             raise UndocumentedSchemaSectionError(
                 UNDOCUMENTED_SCHEMA_SECTION_ERROR.format(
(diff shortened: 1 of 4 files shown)


sondrelg:
Trying to review on a very small screen, so need some help understanding the changes made: do you want the library to support a case like this?

```python
schema_tester = SchemaTester()

def my_test(client):
    response = client.get('api/v1/...')
    assert response.status_code == 204
    schema_tester.validate_response(response=response, ...)
```

will this change allow you to detect a missing response? I.e., will the test fail if response.data is missing and the schema documents an actual response?

Seems like a good change in general 👍

radowit:
If schema documents a response, the if before (that gets the `content` key will fail because response content won't match the spec)

sondrelg:
This looks good @radowit 👍 Would you mind rebasing on master and updating the version in `pyproject.toml` to v2.0.1? I'll release a new version shortly after

sondrelg:
Thanks @radowit!

PR #262, opened 2022-04-14
feat: add custom django testing client

### LINKED ISSUE #261: custom testing API client implementation
Hi :wave:

First of all big thanks for developing this great package!
I have been using `drf-openapi-tester` in a few Django-based projects and it works like a charm, however, usually, I'm adding the same code to every project - [`rest_framework.client.APIClient`](https://github.com/encode/django-rest-framework/blob/master/rest_framework/test.py#L261) subclass that simply validates response against OpenAPI description using `SchemaTester`.
What do you think about adding such a client to `drf-openapi-tester` package?

skarzi:
Introduce `OpenAPIClient` that extends `rest_framework.test.APIClient`
with responses validation against OpenAPI schema by using `SchemaTester`.

Closes #261

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ error.
 
 ## Django testing client
 
-`OpenAPIClient` extends Django REST framework
+The library includes an `OpenAPIClient`, which extends Django REST framework's
 [`APIClient` class](https://www.django-rest-framework.org/api-guide/testing/#apiclient).
 If you wish to validate each response against OpenAPI schema when writing
 unit tests - `OpenAPIClient` is what you need!
@@ -209,7 +209,7 @@ client = OpenAPIClient(schema_tester=schema_tester)
 response = client.get('/api/v1/tests/123/')
 ```
 
-To enforce all developers working on the project to use `OpenAPIClient` simply
+To force all developers working on the project to use `OpenAPIClient` simply
 override the `client` fixture (when using `pytest-django`) or provide custom
 test cases implementation (when using standard Django `unitest`-based approach)
 and then you will be sure all newly implemented views will be validated against
--- a/README.md
+++ b/README.md
@@ -66,9 +66,10 @@ def test_response_documentation(client):
 If you are using the Django testing framework, you can create a base APITestCase that incorporates schema validation:
 
 ```python
-from openapi_tester.schema_tester import SchemaTester
-from rest_framework.test import APITestCase
 from rest_framework.response import Response
+from rest_framework.test import APITestCase
+
+from openapi_tester.schema_tester import SchemaTester
 
 schema_tester = SchemaTester()
 
@@ -200,8 +201,7 @@ If you wish to validate each response against OpenAPI schema when writing
 unit tests - `OpenAPIClient` is what you need!
 
 To use `OpenAPIClient` simply pass `SchemaTester` instance that should be used
-to validate responses and then use it like regular Django testing client
-(TIP: add custom fixture if you are using `pytest` to avoid code boilerplate):
+to validate responses and then use it like regular Django testing client:
 
 ```python
 schema_tester = SchemaTester()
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -192,6 +192,29 @@ When the SchemaTester loads a schema, it runs it through
 specification compliance issues. In case of issues with the schema itself, the validator will raise the appropriate
 error.
 
+## Django testing client
+
+`OpenAPIClient` extends Django REST framework
```
sondrelg, line 197:
```suggestion
The library includes an `OpenAPIClient`, which extends Django REST framework's
```

README.md
```diff
@@ -192,6 +192,29 @@ When the SchemaTester loads a schema, it runs it through
...
+[`APIClient` class](https://www.django-rest-framework.org/api-guide/testing/#apiclient).
+If you wish to validate each response against OpenAPI schema when writing
+unit tests - `OpenAPIClient` is what you need!
+
+To use `OpenAPIClient` simply pass `SchemaTester` instance that should be used
+to validate responses and then use it like regular Django testing client
+(TIP: add custom fixture if you are using `pytest` to avoid code boilerplate):
```
sondrelg, line 204: Instead of the tip here, what do you think about including some example code for both cases? Something like:

> If you're using UnitTest or Django tests, do this ... 

> If you're using pytest do this ...

README.md
```diff
@@ -192,6 +192,29 @@ When the SchemaTester loads a schema, it runs it through
...
+
+```python
+schema_tester = SchemaTester()
+client = OpenAPIClient(schema_tester=schema_tester)
+response = client.get('/api/v1/tests/123/')
+```
+
+To enforce all developers working on the project to use `OpenAPIClient` simply
```
sondrelg, line 212:
```suggestion
To force all developers working on the project to use `OpenAPIClient` simply
```

sondrelg (commented):
Looks great 👍 Have a few comments for the docs, but happy to merge this afterwards.

README.md
skarzi: nice idea, it will be more useful to show developers how to use this client in both cases :+1:

openapi_tester/clients.py
```diff
@@ -0,0 +1,26 @@
...
+if TYPE_CHECKING:
+    from rest_framework.response import Response
+
+    from .schema_tester import SchemaTester
+
+
+class OpenAPIClient(APIClient):
+    """``APIClient`` validating responses against OpenAPI schema."""
+
+    def __init__(self, *args, schema_tester: SchemaTester, **kwargs) -> None:
```
skarzi, line 17: @sondrelg I was thinking about making `schema_tester` optional and adding some method e.g. `_schema_tester_factory` to call when it's `None`, just to make the API more developer-friendly. Please let me know what you think about it.

README.md
```diff
@@ -192,6 +193,64 @@ When the SchemaTester loads a schema, it runs it through
...
+import functools
+
+from django.test.testcases import SimpleTestCase
+from openapi_tester.schema_tester import SchemaTester
+
+schema_tester = SchemaTester()
+
+
+class MyTestCase(SimpleTestCase):
+    client_class = functools.partial(OpenAPIClient, schema_tester=schema_tester)
```
skarzi, line 248: when an idea from comment https://github.com/snok/drf-openapi-tester/pull/262/#discussion_r851448465 will be implemented then this could be simplified to:

```suggestion
from django.test.testcases import SimpleTestCase
from openapi_tester.clients import OpenAPIClient


class MyTestCase(SimpleTestCase):
    client_class = OpenAPIClient
```

openapi_tester/clients.py
sondrelg: Since implementations of the client will only need a single extra line of code (most likely copy pasted from the docs), I'm not sure it's worth the extra effort, but if you want to add it, I'm not opposed to it 🙂

README.md
sondrelg: You could also have 

```
schema_tester = SchemaTester()

class SimpleTestBase(SimpleTestCase):
    client_class = functools.partial(OpenAPIClient, schema_tester=schema_tester)
```

in some test file and just use that instead of `SimpleTestCase` right? That way there's no extra code needed for adding it to your test classes.

skarzi: Exactly, I will add some info that such subclass should be used instead of the original `SimpleTestCase` class

sondrelg:
Please ping me when this is finalised and I'll take another look 👍  Guessing it might already be, but not sure 🙂

sondrelg:
This looks really good @skarzi! Thanks for the effort. I'll release a new minor version shortly 👍

skarzi:
@sondrelg thanks :+1: If you need any help with this project let me know, I am eager to contribute more!

sondrelg:
I don't actively use this library at work currently, so having someone help that actively uses it is super helpful 👍  Not sure if there's a lot to be done right now, but feel free to make suggestions, respond to issues, etc.! 👏

PR #263, opened 2022-04-14
docs: fix link and badges in README.md

skarzi:


sondrelg:
Haha, thanks! Can't believe we didn't notice earlier. Either way, I've just turned off sonarqube for the repo, so removing them seems good 👍

PR #265, opened 2022-06-28
Fix exception when validation against empty list

TheSuperiorStanislav:
When making a request to a list action and getting the empty list, the validator fails because it mistakenly passes an empty dict instead of the original response.


TheSuperiorStanislav (commented):
Dunno why checks for linting falling, because it doesn't complain on my changes

sondrelg:
Thanks @TheSuperiorStanislav!

PR #268, opened 2022-08-26
Update workflows

sondrelg:
Changes include:

- Updates to Poetry v1.2
- Adds dependabot for actions
- Updates pre-commit hooks and fixes new linting errors
- Specifies requirement for `<3.14` for DRF, since drf-yasg is incompatible with this version


.github/workflows/testing.yml
```diff
@@ -10,11 +10,11 @@ jobs:
   linting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
```
skarzi, line 13: You can enable `dependabot` for GitHub Actions, so all actions will be updated automatically.

Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#supported-repositories-and-ecosystems

.github/workflows/testing.yml
```diff
@@ -39,25 +39,27 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ "3.7", "3.8" , "3.9", "3.10" ]
-        django-version: [ "3.0", "3.1", "3.2", "4.0" ]
+        django-version: [ "3.0", "3.1", "3.2", "4.0", "4.1" ]
```
skarzi, line 42: NIP: I would leave here only the newest and LTS versions:

```suggestion
        django-version:
          - "3.2"
          - "4.0"
          - "4.1"
```

Reference: https://www.djangoproject.com/download/ - `Supported Version` section

skarzi: NIP: it will be also nice to include 2 additional tests, that could fail:

+ `python-version: 3.11-dev` and `django-version: "4.1"`
+ `python-version: 3.10` and `django-version` installed from the `main` branch

.github/workflows/testing.yml
```diff
@@ -39,25 +39,27 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: "${{ matrix.python-version }}"
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         id: poetry-cache
         with:
           path: ~/.local
           key: key-3
```
skarzi, line 58: When touching this file, let's update the `key` name to something bringing more context and is dependent on the `python-version` matrix variable (this will speed up also CI pipeline):

```suggestion
          key: "poetry-cache-python${{ matrix.python-version }}"
```

.github/workflows/testing.yml
sondrelg: Yeah true, I use this in a few other projects, and might as well add it here too :+1:

.github/workflows/testing.yml
sondrelg: Is the Poetry installation really Python version specific though? I'm not so sure.

How will it speed up the pipeline?

.github/workflows/testing.yml
sondrelg: I think the main branch is a little too much. Since Django puts out release candidates I think testing those should be enough, unless someone explicitly asks us to check main.

But agree we need to add 4.1, 3.11.0-rc.1 and can remove older v3s :slightly_smiling_face:

.github/workflows/testing.yml
skarzi: `poetry` shouldn't be python version specific, but dependencies could be, that's the reason cache key should be parametrized with python version

.github/workflows/testing.yml
```diff
@@ -38,26 +38,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.7", "3.8" , "3.9", "3.10" ]
-        django-version: [ "3.0", "3.1", "3.2", "4.0" ]
+        python-version: [ "3.7.13", "3.8.13" , "3.9.13", "3.10.6", "3.11.0-rc.1" ]
```
skarzi, line 41: NIP: I think we can go with specifying just the major and minor components of the Python version, so `setup-python` will install the latest Python patch version. Going with this approach (so specifying also patch piece of Python version) we will have to remember to update the `testing.yml` workflow matrix every time a new python patch version will be released.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+    reviewers:
+      - sondrelg
+
+  - package-ecosystem: pip
+    directory: /
+    versioning-strategy: lockfile-only
```
skarzi, line 13: NIP: from my experience, it's nice to have an `auto` versioning strategy when developing and maintaining a package/lib, because it's nice to have some info when a major version of any deps is updated, so we can add support for it in our package

.github/workflows/testing.yml
sondrelg: This cache only caches the poetry installation 🙂

skarzi: Ahh, ok my bad, sorry!
Because `virtualenvs` location-related option (`virtualenvs.in-project
` or `virtualenvs.path`) wasn't set, I thought it will create `.venv` in the default location, but according to docs about [`virtualenvs.create`](https://python-poetry.org/docs/configuration/#virtualenvscreate) poetry will install dependencies in currently active env

Issue #282, opened 2022-11-03
Support for retrieving OpenAPI schema from URL


maticardenas:
I'm currently trying to use the package in our Django test suite for an specific service, the thing is that we have all our schemas in a separated repository that's why we spin up that repository and we can access  the schemas the different schemas for example like  `http://schema_repository:{port}/xyz_schemas/openapi.yaml`.

When attempting to pass that path to `SchemaTester` I found out that it fails to find the file, and checking into the code I see `load_schema` only attempts to read from a file (https://github.com/snok/drf-openapi-tester/blob/ef249511dbe8a526f9745bad855632c710613bf0/openapi_tester/loaders.py#L256) (which makes us now to grab the files we need, store them, pass the path and remove them).

It would be ideal if `load_schema` could support also the possibility of retrieving the schema itself from an URL.

sondrelg:
The library has a concept of "loaders". There is a file-loader, a `drf-yasg`-loader, and a `drf-spectacular`-loader. This sounds like a use-case for an "url-loader" or something along those lines. 

In other words, I think it would make sense to write your own class to handle the case where you need to retrieve a schema from a URL. If you do, a PR is welcome; this sounds useful :+1:

[Here](https://github.com/snok/drf-openapi-tester/blob/master/openapi_tester/loaders.py#L240)'s the `StaticSchemaLoader` for reference. All you would need to do is write your own `load_schema` method which fetches the schema with a GET request.

maticardenas:
@sondrelg thanks for clear guidance! I will submit a PR with this feature probably tomorrow then 👍

PR #283, opened 2022-11-04
feat: adding url schema loader functionality

maticardenas:
Adding functionality for loading a schema from a `url` path.

This can be useful for situations where you need to load local tests and you hold schemas in a separate repository / server.

- Adding `UrlStaticSchemaLoader` to handle the cast of the schema content.
- Checking if `schema_file_path` is an `url` and assigning corresponding loader accordingly.
- Adding tests.

This is related to issue #282

### DIFF
--- a/openapi_tester/loaders.py
+++ b/openapi_tester/loaders.py
@@ -9,6 +9,7 @@ from json import dumps, loads
 from typing import TYPE_CHECKING, cast
 from urllib.parse import urlparse
 
+import httpx
 import yaml
 from django.urls import Resolver404, resolve
 from django.utils.functional import cached_property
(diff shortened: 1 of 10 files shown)


openapi_tester/utils.py
```diff
@@ -58,3 +59,17 @@ def lazy_combinations(options_list: Sequence[dict[str, Any]]) -> Iterator[dict]:
     for i in range(2, len(options_list) + 1):
         for combination in combinations(options_list, i):
             yield merge_objects(combination)
+
+
+def is_path_an_url(path: str) -> bool:
```
sondrelg, line 64: Think we could use the django url validator directly: https://github.com/django/django/blob/stable/1.3.x/django/core/validators.py#L45

openapi_tester/loaders.py
```diff
@@ -258,3 +259,28 @@ def load_schema(self) -> dict[str, Any]:
...
+        self.url = url
+
+    def load_schema(self) -> dict[str, Any]:
+        """
+        Loads a static OpenAPI schema from url, and parses it to a python dict.
+
+        :return: Schema contents as a dict
+        :raises: ImproperlyConfigured
+        """
+        response = httpx.get(self.url)
```
sondrelg, line 280: Better to use a builtin library like requests

sondrelg (changes requested):
Looks good, but I'd rather not pull in httpx as a dependency for this. Better do use a builtin library like urllib or requests. Otherwise we're forcing users to potentially pull in an extra dependency for no real gain.

openapi_tester/utils.py
maticardenas: Nice, I hadn't considered it. Changed it to use django's validator.

openapi_tester/loaders.py
maticardenas: Replaced `httpx` with `requests`

sondrelg:
Looks great @maticardenas 👏

PR #300, opened 2023-07-03
Check anyOf and oneOf schemas in nullable validation

arttuperala:
`test_is_nullable()` is executed before the normal `test_schema_section()` processing, including the `anyOf` and `oneOf` checking, takes place, and if your `nullable` definition comes from inside either `anyOf` or `oneOf`, the validation will fail.

### DIFF
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -239,6 +239,20 @@ class SchemaTester:
         """
         openapi_schema_3_nullable = "nullable"
         swagger_2_nullable = "x-nullable"
+        if "oneOf" in schema_item:
+            one_of: list[dict[str, Any]] = schema_item.get("oneOf", [])
+            return any(
+                nullable_key in schema and schema[nullable_key]
+                for schema in one_of
+                for nullable_key in [openapi_schema_3_nullable, swagger_2_nullable]
+            )
+        if "anyOf" in schema_item:
+            any_of: list[dict[str, Any]] = schema_item.get("anyOf", [])
+            return any(
+                nullable_key in schema and schema[nullable_key]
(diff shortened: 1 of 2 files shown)


arttuperala:
It basically just allows the `test_is_nullable()` to pass without issues if the `nullable: true` comes from within an `anyOf` or `oneOf`.

We're basically working with a JSON:API schema that looks a bit something like this:

```yaml
components:
  schemas:
    reltoone:
      type: object
      properties:
        data:
          $ref: '#/components/schemas/relationshipToOne'
    relationshipToOne:
      anyOf:
      - $ref: '#/components/schemas/nulltype'
      - $ref: '#/components/schemas/linkage'
    nulltype:
      type: object
      nullable: true
      default: null
    linkage:
      type: object
      properties:
        type:
          $ref: '#/components/schemas/type'
        id:
          $ref: '#/components/schemas/id'
        meta:
          $ref: '#/components/schemas/meta'
    ABC:
      type: object
      additionalProperties: false
      properties:
        relationships:
          type: object
          properties:
            xyz:
              $ref: '#/components/schemas/reltoone'

```

And then we'd feed the validator a response like this:

```json
{
    "data": {
        "relationships": {
            "xyz": {
                "data": null
            }
        }
    }
}
```

Even though `/data/relationships/xyz/data` is a `relationshipToOne`, where one of its possible schemas is `nulltype` with `nullable: true`, it'd fail the validation because the current code could not peek inside the `anyOf` array to figure out if one of the possible values is nullable.

sondrelg:
Looks good to me! I'll release a fix shortly 👍

arttuperala:
I might have another PR coming up also, since there's also another issue that I'm having with the validators.

PR #301, opened 2023-07-04
Handle dicts in validate_unique_items

arttuperala:
`validate_unique_items()` validator had a known issue where it could not validate `dict` objects in any way. Attempting to pass one would result in an immediate `TypeError: unhashable type: 'dict'`.

This merge request adds support for validating `dict` objects by serializing them to ordered JSON strings and doing the comparison with those. I imagine that this solution might not be flexible enough to handle every possible `dict` but it should be better than not being able to handle any `dict`. And since we are talking about data returned by an API, it's quite likely that the data passed to the validator will actually be JSON serializable.

### DIFF
--- a/openapi_tester/validators.py
+++ b/openapi_tester/validators.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import base64
+import json
 import re
 from typing import TYPE_CHECKING
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


sondrelg:
Looks good! Thanks @arttuperala 👏

Issue #291, opened 2022-12-01
requestBody schema testing


darduf:
We noticed in our OAS testing that the `requestBody` schema is not tested. For example if the `post.requestBody` is removed from a path (that expects a requestBody) in the OAS, there is no schema checking here and tests pass. Similarly if the `requestBody` is incorrect, the tests also still pass.

Is there scope to include such a test in this package?

Thanks!

sondrelg:
Yes, that sounds like it should happen. Since I'm no longer using this package myself, would you want to help maintain the package doing forward @darduf? I could give you push access, and I could help review changes, but let you add features like this, if you're interested?

darduf:
Thanks for the offer @sondrelg. I don't currently have the bandwidth to maintain the package, but would be open to contribute to the package as previously done on fixes. 
>Since I'm no longer using this package myself, 

Out of interest, what are you using for this type of OAS testing?

sondrelg:
I'm no longer working on an API first Django backend, otherwise I likely would be 👍 Otherwise I think schemathesis might provide similar benefits

maticardenas:
Hi @sondrelg @darduf,  as I'm currently using this for my project I also found useful to have `requestBody` validated within the client. 
I decided to add the feature and [created a PR](https://github.com/snok/drf-openapi-tester/pull/304) for it, would be great if you could give it a check :) 

Thanks!

Issue #303, opened 2023-08-04
Archiving drf-openapi-tester


sondrelg:
It's now been a few years since I've used this personally, so I think I will archive it unless someone else wants to maintain it. Alternative packages like [schemathesis](https://github.com/schemathesis/schemathesis) have had a lot of development since this was created, so think those generally provide more value for less effort (though it's been a while since I've looked at it). 

Opening this in case someone does want to take over maintenance.

@Goldziher, any objections? :relaxed:

PR #304, opened 2023-10-11
feat: adding support for request validation

maticardenas:
issue: https://github.com/snok/drf-openapi-tester/issues/291

# Adding `requestBody` validation support.  

Currently, the package offers the possibility of validating only the response, however, it is useful to also at the same time,e to have validated that the passed `resquestBody` in our functional tests are also properly documented and match the implementation.

## Additions/Changes

- Methods for validating request (`validate_request`) and extract the corresponding `requestBody` schema section from the OpenAPI schema (`get_request_body_schema_section`)
- Updating `test_project` to also expose `Pets` API, as it is the OpenAPI schema we have in tests.
- Adding tests for schema tester, clients and validators.
- Updating affected variables and functions.

## Notes

- Adding support **only** for `openapi` schemas (no `swagger`, previous versions), which involves `StaticSchemaLoader` and `UrlStaticSchemaLoader` instantiated with an `openapi 3.x.x`.  This because it's the current requirement I have for my project.
- Decided to make `request` validation **only** when the scenarios **has a successful** response. Why? In the case you provide for example an incomplete payload, it will always fail naturally when comparing when your schema - in this situations we should let the functional case validate that the test we chose doesn't fit our design.

## Follow Ups

- Add support for request bodies for `swagger`, `DrfSpectacularSchemaLoader` and `DrfYasgSchemaLoader`.
- Add support maybe for checking more aspects of the request? `query` params, `headers`, `authentication`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ from openapi_tester import SchemaTester
 schema_tester = SchemaTester(schema_file_path="./schemas/publishedSpecs.yaml")
 ```
 
-Once you've instantiated a tester, you can use it to test responses:
+Once you've instantiated a tester, you can use it to test responses and request bodies:
 
 ```python
 from openapi_tester.schema_tester import SchemaTester
@@ -53,6 +53,12 @@ def test_response_documentation(client):
     response = client.get('api/v1/test/1')
     assert response.status_code == 200
     schema_tester.validate_response(response=response)
+
+
+def test_request_documentation(client):
+    response = client.get('api/v1/test/1')
+    assert response.status_code == 200
+    schema_tester.validate_request(response=response)
 ```
 
 If you are using the Django testing framework, you can create a base `APITestCase` that incorporates schema validation:
@@ -188,11 +194,11 @@ In case of issues with the schema itself, the validator will raise the appropria
 
 The library includes an `OpenAPIClient`, which extends Django REST framework's
 [`APIClient` class](https://www.django-rest-framework.org/api-guide/testing/#apiclient).
-If you wish to validate each response against OpenAPI schema when writing
+If you wish to validate each request and response against OpenAPI schema when writing
 unit tests - `OpenAPIClient` is what you need!
 
 To use `OpenAPIClient` simply pass `SchemaTester` instance that should be used
-to validate responses and then use it like regular Django testing client:
+to validate requests and responses and then use it like regular Django testing client:
 
 ```python
 schema_tester = SchemaTester()
--- a/openapi_tester/clients.py
+++ b/openapi_tester/clients.py
@@ -27,8 +27,14 @@ class OpenAPIClient(APIClient):
     def request(self, **kwargs) -> Response:  # type: ignore[override]
         """Validate fetched response against given OpenAPI schema."""
         response = super().request(**kwargs)
+        if self._is_successful_response(response):
+            self.schema_tester.validate_request(response)
         self.schema_tester.validate_response(response)
         return response
+
+    @staticmethod
+    def _is_successful_response(response: Response) -> bool:
+        return response.status_code < 400
 
     @staticmethod
     def _schema_tester_factory() -> SchemaTester:
--- a/openapi_tester/schema_tester.py
+++ b/openapi_tester/schema_tester.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import re
+from dataclasses import dataclass
 from itertools import chain
 from typing import TYPE_CHECKING, Any, Callable, cast
 
@@ -13,8 +14,8 @@ from openapi_tester.constants import (
     INIT_ERROR,
     UNDOCUMENTED_SCHEMA_SECTION_ERROR,
     VALIDATE_ANY_OF_ERROR,
-    VALIDATE_EXCESS_RESPONSE_KEY_ERROR,
-    VALIDATE_MISSING_RESPONSE_KEY_ERROR,
+    VALIDATE_EXCESS_KEY_ERROR,
+    VALIDATE_MISSING_KEY_ERROR,
     VALIDATE_NONE_ERROR,
     VALIDATE_ONE_OF_ERROR,
     VALIDATE_WRITE_ONLY_RESPONSE_KEY_ERROR,
@@ -48,6 +49,17 @@ if TYPE_CHECKING:
     from typing import Optional
 
     from rest_framework.response import Response
+
+
+@dataclass
+class OpenAPITestConfig:
+    """Configuration dataclass for schema section test."""
+
+    case_tester: Callable[[str], None] | None = None
+    ignore_case: list[str] | None = None
+    validators: list[Callable[[dict[str, Any], Any], str | None]] | None = None
+    reference: str = "init"
+    http_message: str = "response"
 
 
 class SchemaTester:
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ from openapi_tester import SchemaTester
 schema_tester = SchemaTester(schema_file_path="./schemas/publishedSpecs.yaml")
 ```
 
-Once you've instantiated a tester, you can use it to test responses:
+Once you've instantiated a tester, you can use it to test responses and requests (on successful responses):
```
sondrelg, line 44: Don't think it's necessary to explain this detail here :+1: 
```suggestion
Once you've instantiated a tester, you can use it to test responses and request bodies
```

README.md
```diff
@@ -203,6 +209,7 @@ response = client.get('/api/v1/tests/123/')
 To force all developers working on the project to use `OpenAPIClient` simply
 override the `client` fixture (when using `pytest` with `pytest-django`):
 
+
```
sondrelg, line 212:
```suggestion
```

openapi_tester/clients.py
```diff
@@ -22,14 +22,21 @@ def __init__(
     ) -> None:
         """Initialize ``OpenAPIClient`` instance."""
         super().__init__(*args, **kwargs)
+        #
```
sondrelg, line 25:
```suggestion
```

openapi_tester/clients.py
```diff
@@ -22,14 +22,21 @@ def __init__(
...
         self.schema_tester = schema_tester or self._schema_tester_factory()
 
     def request(self, **kwargs) -> Response:  # type: ignore[override]
         """Validate fetched response against given OpenAPI schema."""
         response = super().request(**kwargs)
+        if self._is_successfull_response(response):
```
sondrelg, line 31: tiny typo here :slightly_smiling_face: 
```suggestion
        if self._is_successful_response(response):
```

openapi_tester/clients.py
```diff
@@ -22,14 +22,21 @@ def __init__(
...
+            self.schema_tester.validate_request(response)
         self.schema_tester.validate_response(response)
         return response
 
+    @staticmethod
+    def _is_successfull_response(response: Response) -> bool:
+        return response.status_code in range(200, 300) and response.status_code != 204
```
sondrelg, line 38: Is there a reason we cannot use `response.ok` or similar built-in functions in the requests response class?

sondrelg: The `response.ok` returns `True` for anything < 400, but seems appropriate here? https://github.com/psf/requests/blob/main/src/requests/models.py#L755

openapi_tester/schema_tester.py
```diff
@@ -338,6 +397,7 @@ def test_openapi_object(
         reference: str,
         case_tester: Callable[[str], None] | None = None,
         ignore_case: list[str] | None = None,
+        **kwargs: Any,
```
sondrelg, line 400: What's this for? Can we use explicit arguments instead?

openapi_tester/schema_tester.py
```diff
@@ -198,6 +199,61 @@ def get_response_schema_section(self, response: Response) -> dict[str, Any]:
             )
         return {}
 
+    def get_request_body_schema_section(self, request: dict[str, Any]) -> dict[str, Any]:
```
sondrelg, line 202: Isn't this type wrong? We're passing in `response.request` further down, and I believe that's why you needed the `type: ignore`'s
```suggestion
    def get_request_body_schema_section(self, request: Request) -> dict[str, Any]:
```

sondrelg (commented):
Looks pretty good! Thanks for taking the time :clap: 

I have a few questions and comments, but happy to accept this in general :+1:

openapi_tester/schema_tester.py
maticardenas: Not sure what I had left it like this, it can certainly be an explicit argument. Changed it, thanks!

openapi_tester/schema_tester.py
maticardenas: Indeed, that was the case :) set proper type and removed the unnecessary `# type: ignore`

openapi_tester/clients.py
maticardenas: I tried this, and this is true for `requests` library `Response` type, but in this case, with `OpenAPIClient` the response we will get is of type `rest_framework.response.Response`. Unfortunately it doesn't count with a built in `ok` property, `raise_for_status` or similar 😢 .  I have modified the check though to consider everything `< 400` as successful.

README.md
maticardenas: Done 👍

README.md
maticardenas: Done 👍

openapi_tester/clients.py
maticardenas: Done 👍

openapi_tester/clients.py
maticardenas: Done 👍

openapi_tester/schema_tester.py
maticardenas: Ok, I think I see why I properly had left it like that. After adding one extra explicit argument `pylint` complains that they are now too many (7, while max. is 6).
In order to overcome this introduced a new `OpenAPITestConfig` dataclass and grouped the optional "test config" params within it, guess it could be improved but found it more clear for now.

maticardenas: Hi, @sondrelg taking a look at this again I think this might be sort of a breaking change? Let me know if you'd like me to take a different approach there :)

sondrelg: Sorry for the radio silence here @maticardenas :bow: I've had a busy few weeks. I'll do my best to get this reviewed by the end of the week!

maticardenas: @sondrelg no problem at all! and thanks 🙏 , I appreciate the review :)

sondrelg:
I think this looks pretty good @maticardenas and would probably accept the PR as it currently stands. That said, I've been considering archiving this project (re https://github.com/snok/drf-openapi-tester/issues/303) for a while, since I haven't used it since 2019, and I think I will do that now instead of accepting this, unfortunately.

I'm happy to see someone is using it, and I would suggest you use your own fork of the project going forward. If you want to maintain the fork beyond this PR, I would be happy to link to it in the README here :+1: If you have any questions about package management or forks you need answered, just let me know and I'll be happy to help

maticardenas:
> I think this looks pretty good @maticardenas and would probably accept the PR as it currently stands. That said, I've been considering archiving this project (re #303) for a while, since I haven't used it since 2019, and I think I will do that now instead of accepting this, unfortunately.
> 
> I'm happy to see someone is using it, and I would suggest you use your own fork of the project going forward. If you want to maintain the fork beyond this PR, I would be happy to link to it in the README here 👍 If you have any questions about package management or forks you need answered, just let me know and I'll be happy to help

@sondrelg thanks for the review, and unfortunate to hear that you are planning to archive the repository as I currently use it, but I can understand the reasons.

I have merged [this same PR](https://github.com/maticardenas/drf-openapi-tester/pull/1) in my fork, and plan to maintain the fork beyond the PR (need to add github actions to it now though :D), so it would be nice the link in the README if possible. Thanks again!

sondrelg:
Sounds good @maticardenas 👍 Sorry for not giving this feedback earlier

